PR #89, opened 2017-09-02
feat: warn when the babel preset is not configured

satya164:


src/sheet.js
```diff
@@ -49,6 +49,10 @@ function sheet() {
     throw new Error('Unable to insert stylesheet');
   }
 
+  if (process.env.NODE_ENV !== 'test') {
```
thymikee: you need to test for `process`, afaik browser will throw evaluating this code.

satya164: it doesn't

zamotany: When prevaling the code NODE_ENV won't be `test`

zamotany: you should check for typeof window i guess

satya164: @zamotany this code will only execute in browser or jsdom since there is a check for `document` above

thymikee:
Looks like you need to add a test :D

satya164:
Sorry, I'm not able to add a test because the environment is always `test` on travis even if I assign `process.env.NODE_ENV` to a different value. I'll merge this for now so I can do a release and add a test tomorrow.

thymikee:
Just remove this console.log

src/__tests__/__snapshots__/sheet.browser.spec.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`sheet module (browser) should warn for configuring babel preset 1`] = `"Babel preset for Linaria is not configured. See https://github.com/callstack-io/linaria/blob/master/docs/BABEL_PRESET.md for instructions."`;
```
thymikee: There's a snapshot but no test

PR #91, opened 2017-09-02
chore: fix publishing to npm

satya164:


package.json
```diff
@@ -16,15 +16,17 @@
     "flow": "flow",
     "lint": "eslint .",
     "test": "jest",
-    "prepublish": "yarn run build:clean && yarn run build:transpile && yarn run build:flow",
+    "prepare": "yarn run build:clean && yarn run build:transpile && yarn run build:flow",
```
zamotany: Why prepare instead of prepublish??

satya164: @zamotany `prepublish` is deprecated in `npm@5`

thymikee:
Linter is not happy

PR #99, opened 2017-09-07
website: switch to preact

satya164:
I also tried to do it without preact-compat, but had trouble with flow


website/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
thymikee: `.eslintrc` is deprecated, use `.eslintrc.json` or `.eslintrc.js`

thymikee:
Have you tried `flow-typed create-stub preact`? Also, maybe we should try React 16 RC? It's ~37kB now

satya164:
> Have you tried flow-typed create-stub preact?

It'll only suppress the errors, and the preact code won't typecheck

> Also, maybe we should try React 16 RC? It's ~37kB now

Still much bigger without providing any advantages

thymikee:
If we really want to get rid of example directory, please do so as a separate PR before merging this.

satya164:
I don't want to do multiple PRs and do unnecessary changes to fix errors in example directory. Let's just rebase and merge this.

PR #105, opened 2017-09-09
remove + sign for users simple copy and paste

pronevich:


README.md
```diff
@@ -39,11 +39,11 @@ Add the `linaria/babel` preset to your Babel configuration:
   "presets": [
```
satya164: can you also change `diff` to `json`?

pronevich: done)

PR #109, opened 2017-09-11
refactor: use single implementation of sheet

satya164:
This will allow us to use the same API in browser and node, and will be useful for things like custom properties.


src/sheet.js
```diff
@@ -3,53 +3,22 @@
...
-      dump() {
-        const result = cssText || '';
-        cssText = null;
-        cache = {};
-        return result;
-      },
-    };
-  }
+  const browser =
+    typeof window === 'object' && window != null && window.document != null;
```
thymikee: since it's boolean, `isBrowser` maybe?

src/sheet.js
```diff
@@ -3,53 +3,22 @@
...
+  // eslint-disable-next-line global-require
+  const document = browser ? window.document : require('./document').default;
```
thymikee: No such file?

satya164: What do you mean?

thymikee: Sorry, didn't spot this

screen shot 2017-09-11 at 16 52 14

PR #115, opened 2017-09-22
fix: Append classnames when extracting to single file

thymikee:
**Summary**

When extracting with `{single: true}` option, classnames were overwritten instead of appended.

**Test plan**

Added a test that verifies proper behavior.


src/babel/preval-extract/__tests__/extractStyles.spec.js
```diff
@@ -56,6 +56,37 @@ describe('preval-extract/extractStyles module', () => {
     expect(writeFileSync).not.toHaveBeenCalled();
   });
 
+  it('should append classnames when writing to single file', () => {
```
zamotany: This test case is a duplication of https://github.com/callstack/linaria/blob/a33b10b6cdd8a16ea7b02952624268f5a4872fcf/src/babel/preval-extract/__tests__/extractStyles.spec.js#L223

src/babel/preval-extract/extractStyles.js
```diff
@@ -90,16 +90,19 @@ export default function extractStyles(
...
-    return;
-  }
-
   if (cache) {
-    stylesCache[filename] = data;
+    if (single) {
+      if (hasCachedStyles(currentFilename, data)) {
+        return;
+      }
+      stylesCache[currentFilename] = data;
```
zamotany: Use `filename`

thymikee: that's the issue, filename is always the same and get's overwritten

src/babel/preval-extract/extractStyles.js
```diff
@@ -90,16 +90,19 @@ export default function extractStyles(
...
-      addRequireForCss(types, program, filename);
-    }
-
-    return;
-  }
-
   if (cache) {
-    stylesCache[filename] = data;
+    if (single) {
+      if (hasCachedStyles(currentFilename, data)) {
```
zamotany: Use `filename`

src/babel/preval-extract/__tests__/extractStyles.spec.js
thymikee: not quite, as that other test doesn't test for appending stylesheets, just calling it once.

zamotany: You should update this one

thymikee: kk, will do.

src/babel/preval-extract/extractStyles.js
zamotany: Why `currentFilename` is used instead of `filename`?

src/babel/preval-extract/extractStyles.js
zamotany: Same as above

src/babel/preval-extract/__tests__/extractStyles.spec.js
zamotany: my bad, then move your case to `with cache enabled` describe block

src/babel/preval-extract/extractStyles.js
zamotany: ok

satya164: shouldn't it be `currentFilePath`?

thymikee: Hm, this got me thinking, that if this path is not absolute, we need to process it through `parseCurrentFilename`, otherwise we can overwrite cache of 2 unrelated files which have the same names. I'll post a followup later.

PR #119, opened 2017-09-24
feat: add a hoist-css plugin. resolves #110

### LINKED ISSUE #110: Babel plugin to Hoist CSS
What if we could write a CSS rule inline without having to give it a name? A Babel plugin can automatically hoist this to the program scope and give it a name.

This JavaScript:

```js
function MyComponent() {
  return <article {...styles(css`color: blue;`)} />;
}
```

Gets transpiled into:

```js
function MyComponent() {
  return <article {...styles(_article)} />;
}
var _article = css`color: blue;`;
```

Context: https://twitter.com/ryanflorence/status/907316693084418048

Implementation:

```js
export default function(babel) {
  const { types: t } = babel;

  return {
    visitor: {
      Program: {
        enter(path, state) {
          state.cssRules = {};
        },
        exit(path, state) {
          Object.keys(state.cssRules).forEach(name => {
            path.node.body.push(
              t.variableDeclaration("var", [t.variableDeclarator(t.identifier(name), state.cssRules[name])])
            );
          });
        }
      },
      JSXOpeningElement(path, state) {
        path.node.attributes.forEach(attr => {
          if (t.isJSXSpreadAttribute(attr) && t.isCallExpression(attr.argument) && attr.argument.callee.name === "styles") {
            attr.argument.arguments = attr.argument.arguments.map(arg => {
              if (t.isTaggedTemplateExpression(arg) && arg.tag.name === "css") {
                const className = path.scope.generateUidIdentifier(path.node.name.name);
                state.cssRules[className.name] = arg;
                return className;
              }
              return arg;
            });
          }
        });
      }
    }
  };
}
```

satya164:
Usually we need to think of a className every time we write some style rules. While this makes sense for rules with well-defined styles, there are many instances where we have a one off rule like `flex: 1` etc. It doesn't make sense to come up with a name for such a rule.

The babel plugin will transpile this:

```js
function MyComponent() {
  return <div {...styles(css`flex: 1`)} />;
}
```

to this:

```js
function MyComponent() {
  return <div {...styles(_div)} />;
}
var _div = css`flex: 1;`;
```

The plugin also supports `className` and `class` props (for preact) along with `styles`.

Currently we don't support such usage with our babel plugin even if it works fine in runtime, so it looks like it should work in compile time.

More context: #110

### DIFF
--- a/src/babel/hoist-css/index.js
+++ b/src/babel/hoist-css/index.js
@@ -1,0 +1,64 @@
+/* @flow */
+
+import type { BabelTypes, NodePath } from '../types';
+
+type State = {
+  programPath: NodePath<*>,
+};
+
+const transformTaggedTemplate = (
+  t: BabelTypes,
+  path: NodePath<*>,
+  state: State,
+  expression: *
+) => {
+  if (
+    t.isTaggedTemplateExpression(expression) &&
+    expression.tag.name === 'css'
+  ) {
+    const className = path.scope.generateUidIdentifier(path.node.name.name);
+    state.programPath.node.body.push(
+      t.variableDeclaration('var', [
+        t.variableDeclarator(t.identifier(className.name), expression),
+      ])
+    );
+    return className;
+  }
+  return expression;
+};
+
+export default function({ types: t }: { types: BabelTypes }) {
+  return {
+    visitor: {
+      Program: {
+        enter(path: NodePath<*>, state: State) {
+          state.programPath = path;
+        },
+      },
+      JSXOpeningElement(path: NodePath<*>, state: State) {
+        path.node.attributes.forEach(attr => {
+          if (
+            t.isJSXSpreadAttribute(attr) &&
+            t.isCallExpression(attr.argument) &&
+            attr.argument.callee.name === 'styles'
+          ) {
+            attr.argument.arguments = attr.argument.arguments.map(arg =>
+              transformTaggedTemplate(t, path, state, arg)
+            );
+          } else if (
+            t.isJSXIdentifier(attr.name) &&
+            t.isJSXExpressionContainer(attr.value) &&
+            (attr.name.name === 'class' || attr.name.name === 'className')
+          ) {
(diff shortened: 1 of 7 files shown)


src/babel/hoist-css/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+export default function({ types: t }: { types: BabelTypes }) {
+  return {
+    visitor: {
+      Program: {
+        enter(path: NodePath<*>, state: State) {
+          state.cssRules = {};
+        },
+        exit(path: NodePath<*>, state: State) {
+          Object.keys(state.cssRules).forEach(name => {
```
zamotany: You don't have to do this here. Just save the reference to `Program` path in `enter` and then add new variable declarations in `transformTaggedTemplate`.

src/babel/hoist-css/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+import type { BabelTypes, NodePath, BabelCallExpression } from '../types';
+
+type State = {
+  cssRules: {
+    [className: string]: BabelCallExpression,
+  },
+};
+
+const transformTaggedTemplate = (t, path, arg, state) => {
```
zamotany: Missing flow types.

satya164: @zamotany flow doesn't need me to specify types here since it's internal and is not exported, it can infer the types

src/babel/hoist-css/index.js
satya164: @zamotany you mean in a scoped variable?

src/babel/hoist-css/index.js
zamotany: I know, but it's easier for ppl to understand the code.

src/babel/hoist-css/index.js
zamotany: no. I mean, move this logic to `transformTaggedTemplate` and rewrite it to use `state.program` to add new variables.

PR #123, opened 2017-10-08
css.named validation

### LINKED ISSUE #52: Ensure we throw nice errors when users do something unsupported
### Using undefined variables

Handled in #81

### Errors thrown during preval

Handled in #111 

If a function throws an error during `preval`, we should catch it and throw nice error with proper stacktrace.

```js
const test = () => {
  throw new Error('Some weird error');
}

const test = css`
  color: ${test()};
`;
```

### Typos in CSS

If the user mistyped a variable name, we could print an error instead of generating invalid CSS. For example, say we get `undefined` during interpolation:

```js
const colors = {
  blue: 'blue'
}

cont title = css`
  color: ${colors.bleu};
`
```

zamotany:
Closes #52 (3rd case)

screen shot 2017-10-08 at 12 26 47
screen shot 2017-10-08 at 12 27 32

### DIFF
--- a/src/__tests__/css.spec.js
+++ b/src/__tests__/css.spec.js
@@ -3,7 +3,11 @@
 import css from '../css';
 
 jest.mock('../babel/lib/errorUtils', () => ({
-  getFramesFromStack: (...args) => require.requireActual('../babel/lib/errorUtils').getFramesFromStack(...args),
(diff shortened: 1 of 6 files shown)


zamotany:
Before any1 asks: I'm aware that some tests are missing.

PR #125, opened 2017-10-09
Use latest linaria on website

thymikee:
**Summary**

Use linaria from node_modules, we don't need to link it anymore to website.

**Test plan**

Tested manually regular and server-rendered site


zamotany:
With linked linaria we can test it with actual project. Right now we don't have dedicated E2E test and this website was behaving like this.

PR #126, opened 2017-10-09
chore: provide example config with CRA and Next.js

thymikee:
**Summary**

Describes how to setup Linaria with CRA and Next.js.
I'm ok with moving it somewhere else.

### DIFF
--- a/docs/BABEL_PRESET.md
+++ b/docs/BABEL_PRESET.md
@@ -120,7 +120,7 @@ This will extract your CSS to `src/.linaria-cache` directory (add it to your `.g
 }
 ```
 
-Because `next` is doing server rendering, we'll need to adjust our config properly. We need to extract styles to a single file, which we can then be declared in e.g. custom `<Head>` component. We need to override Next.js's `babel-preset-env` `"modules": false` with `"modules": "commonjs"` (default), because Linaria needs transpiled the modules. You can still leverage Webpack's dynamic imports by using `"modules": false` in `babel-loader` config.
+Since `next` server renders our app, we need to tweak our configuration. We want to extract styles to a single file, which we can then be declared in e.g. custom `<Head>` component. We need to override Next.js's `babel-preset-env` `"modules": false` with `"modules": "commonjs"` (default), because Linaria needs transpiled the modules. You can still leverage Webpack's dynamic imports by using `"modules": false` in `babel-loader` config.
 
 ```js
 import Head from 'next/head';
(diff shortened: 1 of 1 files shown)


docs/BABEL_PRESET.md
```diff
@@ -75,3 +75,69 @@ Remember to set `process.env.BABEL_ENV` or `process.env.NODE_ENV` to `test`. If
...
+        {
+          "outDir": "./src/.linaria-cache"
+        }
+      ]
+    ]
+  }
+}
+```
+
+This will extract your CSS to `src/.linaria-cache` directory (add it to your `.gitignore`). We need to override CRA's `babel-preset-env` `"modules": false` with `"modules": "commonjs"` (default), because Linaria needs transpiled the modules. You can still leverage Webpack's dynamic imports by using `"modules": false` in `babel-loader` config.
```
satya164:
```
Linaria needs to transpile ES2015 modules to require calls, so we need to configure babel to do the transpilation. ...
```

cc @zamotany maybe you want to always transpile requires for linaria so that users don't have to disable this?

docs/BABEL_PRESET.md
```diff
@@ -75,3 +75,69 @@ Remember to set `process.env.BABEL_ENV` or `process.env.NODE_ENV` to `test`. If
...
+        "filename": "style.css",
+        "outDir": "./static",
+        "single": true
+      }
+    ]
+  ]
+}
+```
+
+Because `next` is doing server rendering, we'll need to adjust our config properly. We need to extract styles to a single file, which we can then be declared in e.g. custom `<Head>` component. We need to override Next.js's `babel-preset-env` `"modules": false` with `"modules": "commonjs"` (default), because Linaria needs transpiled the modules. You can still leverage Webpack's dynamic imports by using `"modules": false` in `babel-loader` config.
```
satya164:
```
Since `next` server renders our app, we need to tweak our configuration.
```

docs/BABEL_PRESET.md
zamotany: Linaria needs `module` set to whatever except for false, because it will load this config and use it for transpilation. We can create workaround for this by accepting custom babel config in `linaria` preset options.

thymikee: I'm for this workaround.

satya164: I know. I'm suggesting to always use `es2015-modules-commonjs` regardless of users config.

zamotany: Ah, that might work as well. And it's better actually for this case.

zamotany: Can you add this plugin here: https://github.com/callstack/linaria/blob/master/src/babel/lib/moduleSystem.js#L94-L98??

satya164: ok, will try

thymikee:
Agreed 👍

PR #127, opened 2017-10-10
fix: guard for first frame in buildCodeFrameError

thymikee:
**Summary**

This came out while testing Linaria with CRA and Next.js, for some reason (didn't have time to investigate why) `frames` were passed either empty or with undefined object.

**Test plan**

covered


zamotany:
If you're still interested this might be why it's happening:
https://github.com/callstack/linaria/blob/master/src/babel/lib/errorUtils.js#L22-L25

PR #128, opened 2017-10-10
feat: handle object properties with hoist-css

satya164:


src/babel/hoist-css/__tests__/__snapshots__/index.spec.js.snap
```diff
@@ -36,6 +36,24 @@ var _article = css\`color: \${colors.red};\`;
...
 `;
 
+exports[`hoist-css babel plugin should hoist CSS from styles object 1`] = `
+"
+const styles = {
+  text: _text,
+  container: _container
+};
+
+var _text = css\`
```
thymikee: We need to pull these variables up, so they're available in `styles` object.

satya164:
I'd revisit this later.

PR #135, opened 2017-10-13
chore: add MIT license

thymikee:
**Summary**

Hey, we're missing the license! We should also add a comment with copyright, like we do in Haul

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Callstack
+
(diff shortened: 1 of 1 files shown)


satya164:
> We should also add a comment with copyright

Why do we need a comment?

thymikee:
So that it's preserved when somebody is copying our code (at least when doing so legally).

thymikee:
cc @grabbou – should we add license headers to src files?

grabbou:
I think we can add it. I got used to doc blocks at the top and that just reads nicer to me

PR #136, opened 2017-10-15
chore: setup typescript-styled-plugin for vscode on website

thymikee:
**Summary**

This is just a proof of concept, that Linaria works with `typescript-styled-plugin` without additional config. Someday they'll create extensions, but until then we can use this approach.
Added a VS Code config so it's already set up if someone uses it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "typescript.tsdk": "website/node_modules/typescript/lib"
+}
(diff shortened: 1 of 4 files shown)


zamotany:
Shouldn't `.vscode` folder be nested in `website`??

thymikee:
I'll check later if it works, just didn't try

thymikee:
Unfortunately nesting `.vscode` doesn't work :(

PR #137, opened 2017-10-17
[WIP] feat: add testing utilities with a snapshot serializer

thymikee:
**Summary**

Adds Jest testing utilities alongside with serializer as a separate package: `linaria/jest`. This is a subject of change and discussion on how we want to expose it (can be a standalone package).

What's worrying me is that I couldn't make the serializer work with Babel preset explicitly specified in `.babelrc`, not yet sure why. So for now I'm using Linaria without the preset, so it generates consistent class names like `css__ab12312` (which is also easier to process). 
 
**Test plan**

Added a test for React and Preact components.


thymikee:
I'll move it to separate package

PR #140, opened 2017-10-18
Stylelint preprocessor

### LINKED ISSUE #8: ESLint or Stylelint plugin to lint styles

zamotany:
Closes #8 

We need to figure out some nice stylelint config.


src/__tests__/css.spec.js
```diff
@@ -38,4 +38,22 @@ describe('css module', () => {
...
+    const header = css.named('header', 'test.js')`
+      color: blue;
+    `;
+    process.env.LINARIA_COLLECT_RAW_STYLES = undefined;
+
+    expect(getRawStyles()).toEqual({
+      'test.js': {
+        template: ['\n  color: blue;\n'],
+        expressions: [],
+        classname: header,
```
thymikee: Shouldn't this be explicitly `header__1r77qux`?

zamotany: No, it should be the same classname as the one returned from `css.named`.

thymikee: ok!

thymikee:
Can you also update the lockfile?

stylelint-config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  extends: 'stylelint-config-recommended',
+  rules: {
+    'at-rule-name-case': 'lower',
+    'unit-case': 'lower',
+    'at-rule-semicolon-newline-after': 'always',
+    'max-empty-lines': 1,
+    'no-empty-source': null,
+    'length-zero-no-unit': true,
```
satya164: isn't this a style thing? 😬

src/css.js
```diff
@@ -5,6 +5,12 @@ import slugify from './slugify';
 import sheet from './sheet';
 import { getFramesFromStack, enhanceFrames } from './babel/lib/errorUtils';
 
+const rawStyles = {};
```
satya164: since we usually do the caching stuff inside `sheet`, I'd suggest to move this there

src/tools/stylelint-preprocessor.js
```diff
@@ -0,0 +1,97 @@
...
+function toString(templates: string[], expressions: string[]) {
+  return templates.reduce(
+    (acc, template, i) =>
+      `${acc}${template}${i >= expressions.length ? '' : expressions[i]}`,
+    ''
+  );
+}
+
+export default function linariaStylelintPreprocessor(/* options */) {
+  process.env.LINARIA_OVERWRITE_BABEL_PRESET = JSON.stringify({
```
satya164: maybe call this `LINARIA_BABEL_PRESET_OPTIONS` or `LINARIA_BABEL_PRESET_OVERRIDES`

stylelint-config.js
satya164: let's remove style rules such as `max-empty-lines`, `at-rule-semicolon-newline-after`, `length-zero-no-unit` and provide a base config which caches only errors. style rules will be addressed by prettier anyway.

website/.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "processors": ["linaria/stylelint-preprocessor"],
```
satya164: can `processors` and `syntax` be inside our config? does it make sense?

src/css.js
zamotany: No, styles in sheet are already processed.

satya164: I mean to add a new variable for unprocessed styles in sheet instead of in `css`

.circleci/config.yml
```diff
@@ -60,13 +60,13 @@ jobs:
...
-  build-website:
+  build-and-lint-website:
       <<: *defaults
       steps:
         - attach_workspace:
             at: ~/linaria
-        - run: |
-            cd website && yarn run build:all && cd ..
+        - run: cd website && yarn run build:all && cd ..
+        - run: cd website && yarn run lint-css && cd ..
```
satya164: maybe we should do

```
- run : |
  cd website
  yarn run build:all
  yarn run lint-css
  cd ..
```

instead?

.eslintignore
```diff
@@ -4,3 +4,4 @@ coverage/
 flow-typed/
 node_modules/
 vendor/
+website/
```
satya164: we shouldn't ignore website from linting

zamotany: we should otherwise it will throw some weird errors

zamotany:
```
✖ 551 problems (494 errors, 57 warnings)
```
all of them from website

PR #141, opened 2017-10-18
add playground link

Tushkiz:
**Summary**

Add link to CSS in JS playground

**Test plan**
NA


zamotany:
Also can you use `###` (h3) instead of `##` (h2), since it has a horizontal line underneath, which doesn't look good for a link.

PR #144, opened 2017-10-19
chore: split integration tests for speed

thymikee:
**Summary**

shaves off 2x time when running integration tests 

**Test plan**

Tests:       36 passed, 36 total
Snapshots:   37 passed, 37 total


setupJest.js
```diff
@@ -0,0 +1,20 @@
...
+import escapeStringRegexp from 'escape-string-regexp'; // eslint-disable-line import/no-extraneous-dependencies
+
+/* istanbul ignore next */
+const serialize = val =>
+  stripAnsi(
+    val
+      .toString()
+      .replace(
+        new RegExp(escapeStringRegexp(process.cwd()), 'gm'),
+        '<<REPLACED>>'
```
zamotany: What about `CWD`??

thymikee: I'll replace it, good call

Issue #138, opened 2017-10-17
Assign css template literals to an object's key


berbaquero:
**Do you want to request a *feature* or report a *bug*?**
Request feature.

**What is the current behavior?**
Linaria only allows to assign `css` template literals to a variable: 

```js
// works nicely
const style = css`
  color: white;
  display: inline-block;
`;
```

The babel plugin fails when assigning to an object's key:

```js
// fails
const style = {
  wrapper: css`
    color: black;
  `,
};
```

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
...

**What is the expected behavior?**

An alternative convention is to declare all the styles of a component in a single object:

```jsx
const styles = {
  root: css`
    color: white;
  `,
  header: css`
    padding: 3em;
  `,
}

<div className={styles.root}>
  <header className={styles.header}>
```

**Please provide your exact Babel configuration and mention your Linaria, Node, Yarn/npm version and operating system.**
...

satya164:
We're working on supporting this :)

zamotany:
The bug should be fixed by #134

PR #146, opened 2017-10-20
feat: support template literals in object properties and jsx props

### LINKED ISSUE #134: Merge hoist-css plugin with preval-extract

satya164:
fixes #134, #138


src/babel/preval-extract/index.js
```diff
@@ -90,17 +89,57 @@ export default (babel: BabelCore) => {
...
+            const parent = path.findParent(
+              p => p.isObjectProperty() || p.isJSXOpeningElement()
+            );
+
+            if (parent) {
+              if (parent.isJSXOpeningElement()) {
+                title = parent.node.name.name;
+              } else {
+                title = parent.node.key.name;
+              }
```
zamotany: You should generate unique id after this line, otherwise the following will generate the same classnames:
```js
const a = { header: css`...` }
const b = { header: css`...` }

// or

<article className={css`...`} />
<div>
  <article className={css`...`} />
</div>
```

Maybe `title = path.scope.generateUidIdentifier(title).name`.

zamotany: This will become a major problem if we settle on using `filename` in production also (#147).

satya164: @zamotany will the unique id be consistent across builds? also see https://github.com/callstack/linaria/issues/147

zamotany: yes

PR #148, opened 2017-10-21
tests: add basic test with puppeteer for the website

satya164:
Depends on #150


satya164:
Please rebase and merge while merging. Thanks :)

PR #154, opened 2017-10-23
Fix benchmark script

### LINKED ISSUE #153: Benchmark script is broken
Running the benchmark script produces `0.00` for everything.

zamotany:
Fixes #153

### DIFF
--- a/benchmark/run.js
+++ b/benchmark/run.js
@@ -12,17 +12,18 @@ const resultsTable = new Table({
   colWidths: [60, 10, 10],
 });
 
-const outDir = path.join(__dirname, 'suites/_temp');
+const outDir = 'benchmark/_temp';
 fs.mkdirSync(outDir);
 
 function transpile(filename, opts = {}) {
-  babel.transformFileSync(path.join(__dirname, filename), {
-    presets: ['env', 'stage-2'],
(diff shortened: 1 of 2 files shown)


benchmark/run.js
```diff
@@ -12,17 +12,18 @@ const resultsTable = new Table({
   colWidths: [60, 10, 10],
 });
 
-const outDir = path.join(__dirname, 'suites/_temp');
+const outDir = 'benchmark/_temp';
```
satya164: maybe `path.join(__dirname, '_temp');` ?

satya164:
Maybe we should add a snapshot test for the benchmark.
- When someone makes changes, they'll be aware of the perf impact
- Will make sure that the script keeps working

benchmark/run.js
zamotany: `outDir` must be __relative__ to CWD

zamotany:
Adding snapshot test doesn't make any sense, those are timings which __changes__ between runs even on the same machine. Manually storing results of benchmark in JSON files and comparing it with some threshold would be a acceptable solution.

If you want to ensure the benchmark is working, we can run it on Circle CI.

PR #173, opened 2017-11-10
tests: fix website integration tests

satya164:
not sure why they broke, but probably because of updates to the fonts


thymikee:
Screenshot diff suggests that the font is no longer Montserrat, while it should be (maybe puppeteer doesn't wait for google font to download).

satya164:
> Screenshot diff suggests that the font is no longer Montserrat

Looks like Montserrat to me, but thinner

thymikee:
But we use the bold one for hero... https://github.com/callstack/linaria/blob/0fde9fd79c185a3451a93c64b8a0e925afc141a5/website/src/components/Hero.js#L35.

satya164:
> But we use the bold one for hero

All of them look thinner in the new snapshot, both body and header ones.

thymikee:
Do they look like this locally? If so, I'm fine merging it.

satya164:
Yea

PR #176, opened 2017-11-20
Support JSON files and create dummy wrapper module for other non-JS extensions

### LINKED ISSUE #159: Not compatible with Webpack file and url loaders?
Bug.

Webpack’s `file-loader` or `url-loader` (they come standard with `create-react-app`) let you `require()` (or `import`) any file type (like images and fonts). At build time, the `require` statement gets replaced with a URL that you can use to load the file at runtime. Unfortunately Linaria seems to break any time I reference anything that gets loaded this way. It appears to attempt to load the file as if it were a JS module. E.g. both of these fail:

```js
css`background-image: url(`require('image.svg')`);`
css`@font-face { font-family: Test; src: url(${require('webfont.woff')})}`;
```
**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
Quick `create-react-app` that attempts to load an SVG and a web font on the home page: https://github.com/steadicat/linaria-test

Expected behavior is for Linaria to let the loader run first so that those `require`s become plains strings to be included in the CSS.

**Please provide your exact Babel configuration and mention your Linaria, Node, Yarn/npm version and operating system.**
See app above.

zamotany:
Fixes #159

### DIFF
--- a/src/babel/lib/moduleSystem.js
+++ b/src/babel/lib/moduleSystem.js
@@ -222,7 +222,7 @@ function getRequireMock(parent: ?Module) {
      * For non JS/JSON requires, we create a dummy wrapper module and just export
      * the moduleId from it, thus letting the bundler handle the rest.
      */
-    if (/\.(?!\/)(?!js$|json$).+$/.test(moduleId)) {
+    if (/\.(?!js)[a-zA-Z0-9]+$/.test(moduleId)) {
       return instantiateModule(
         `module.exports = '${moduleId}'`,
         moduleId,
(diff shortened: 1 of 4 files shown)


src/babel/lib/moduleSystem.js
```diff
@@ -90,6 +90,16 @@ export class Module {
   _compile(code: string, altFilename?: string): Exports {
     const filename = altFilename || this.filename;
 
+    if (/\.json$/.test(filename)) {
+      try {
+        this.exports = JSON.parse(code);
+      } catch (error) {
+        throw new Error(`${error.message} (${filename})`);
```
thymikee: If we want to be super nice, we can use https://github.com/zaach/jsonlint somewhere in the future.

PR #179, opened 2017-11-24
chore: Update eslint-config-callstack-io

thymikee:
**Summary**

Bumping deps

**Test plan**

Website snapshot got updated, not sure if that's because I have Montserrat installed as system font


satya164:
> Website snapshot got updated, not sure if that's because I have Montserrat installed as system font

You likely have an older version. Download the latest version from Google fonts.

thymikee:
But regardless my local fonts, Google version should be displayed, so that's weird.

satya164:
Can you revert the snapshot changes? It didn't change, so the commiting the file is unnecessary.

Issue #175, opened 2017-11-14
Minify class names in production


steadicat:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
In production, generated class names are of the form `_{variableName}__{hash}`, same as in development mode.

**What is the expected behavior?**
They should just be `_{hash}`. This will save some bytes and look cleaner, by not leaking internal variable names into the generated markup and CSS. At least make this an option.

zamotany:
For assignee:
This should be configurable as a babel preset options like `minifyClassnames` and should generate short hash (6 chars should be fine) from `_{variableName}__{hash}` to `ln{hash}` where `ln` is our prefix.

The code should probably be done here https://github.com/callstack/linaria/blob/master/src/babel/preval-extract/prevalStyles.js#L63

PR #183, opened 2017-12-04
Minify classnames

jukben:
This PR added option to minify class names. Closes https://github.com/callstack/linaria/issues/175

Added corresponding test.

You can use it like
```js
{
  "presets": [
    "env",
    "react",
    ["linaria/babel", {
      "single": true,
      "filename": "styles.css",
      "outDir": "dist",
      "minifyClassnames": true,
    }]
  ]
}
```

will produce classs names within pattern `/ln[a-z0-9]{6}/`


src/babel/preval-extract/__tests__/prevalStyles.spec.js
```diff
@@ -76,4 +87,18 @@ describe('preval-extract/prevalStyles', () => {
       "css.named('header', 'filename.js')`color: #ffffff`"
     );
   });
+
+  it('test', () => {
```
thymikee: nit: `it('minifies class names', `

src/babel/preval-extract/__tests__/prevalStyles.spec.js
```diff
@@ -76,4 +87,18 @@ describe('preval-extract/prevalStyles', () => {
...
+      value: className,
+    } = runAssertions(
+      "css.named('header', 'test.js')`color: #ffffff`",
+      'css`color: #ffffff`',
+      {
+        minifyClassnames: true,
+      }
+    );
+
+    expect(className.startsWith('ln')).toBeTruthy();
```
thymikee: You could also assert on the digits count

src/babel/preval-extract/prevalStyles.js
```diff
@@ -11,13 +11,18 @@ import type {
...
+import slugify from '../../slugify';
 
 import getReplacement from './getReplacement';
 import {
   instantiateModule,
   clearLocalModulesFromCache,
 } from '../lib/moduleSystem';
 
+function getMinifiedClassName(className: string) {
+  return `ln${slugify(className).substr(0, 6)}`;
```
thymikee: I'm afraid this may lead to clashing hashes

jukben: Me too, but I'm not capable to calculate in which case. Should I introduce some algorithm which targets only 6 chars long string to avoid substr?

src/babel/preval-extract/__tests__/prevalStyles.spec.js
jukben: Good call!

src/babel/preval-extract/__tests__/prevalStyles.spec.js
jukben: Sorry about that, I've noticed it just now 🙏

src/babel/preval-extract/prevalStyles.js
thymikee: I think it's safer to stick with our 7-char hash, at least we stay consistent with dev vs prod.

jukben: What do you think @zamotany?

src/babel/preval-extract/__tests__/prevalStyles.spec.js
```diff
@@ -76,4 +87,19 @@ describe('preval-extract/prevalStyles', () => {
...
+    } = runAssertions(
+      "css.named('header', 'test.js')`color: #ffffff`",
+      'css`color: #ffffff`',
+      {
+        minifyClassnames: true,
+      }
+    );
+
+    expect(/ln[\w]{6}/.test(className)).toBeTruthy();
+    expect(className.startsWith('ln')).toBeTruthy();
```
satya164: I think this assertion is unnecessary when the previous one already tests this.

src/babel/preval-extract/prevalStyles.js
jukben: @thymikee I'm lame, I don't get it where I got this code. Slugify return 6 chars hash implicitly. Substr is not needed so I assume everything will be OK. 🤷‍♀️

src/babel/preval-extract/prevalStyles.js
```diff
@@ -11,13 +11,18 @@ import type {
...
+import slugify from '../../slugify';
 
 import getReplacement from './getReplacement';
 import {
   instantiateModule,
   clearLocalModulesFromCache,
 } from '../lib/moduleSystem';
 
+function getMinifiedClassName(className: string) {
+  return `ln${slugify(className)}`;
```
satya164: Not really sure if slugifying an already slugified string is a good idea. I'd prefer to change https://github.com/callstack/linaria/blob/master/src/babel/preval-extract/index.js#L111 to set the title to `ln` and work from there.

src/babel/preval-extract/__tests__/prevalStyles.spec.js
```diff
@@ -76,4 +87,19 @@ describe('preval-extract/prevalStyles', () => {
...
+      value: className,
+    } = runAssertions(
+      "css.named('header', 'test.js')`color: #ffffff`",
+      'css`color: #ffffff`',
+      {
+        minifyClassnames: true,
+      }
+    );
+
+    expect(/ln[\w]{6}/.test(className)).toBeTruthy();
```
satya164: This regex should be `[a-zA-Z0-9]`

jukben: You are right, I forget that \w match also `_`. 👶

src/babel/preval-extract/prevalStyles.js
zamotany: Just replace `slugify` with `short-hash`

thymikee:
@jukben please don't run prettier over package.json

jukben:
It looks like I have to turn off prettier for Linaria package, sorry about that.

Issue #169, opened 2017-11-10
Deleting outDir breaks Linaria for good


steadicat:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

I sometimes see stale generated stylesheets (more on that perhaps in future issues), so I find myself deleting `./src/.linaria-cache` (my `outDir`) to try to start fresh. After I do this, Linaria fails to build, spewing a bunch of errors like:

```
`Error in ./src/ui/components/chrome/SiteServicesMenu.js
Module not found: [CaseSensitivePathsPlugin] `src/.linaria-cache/src/ui/components/chrome/SiteServicesMenu.css` does not match the corresponding path on disk - File does not exist.
**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
```

Restarting Webpack or clearing the Babel cache does not help. I'm stuck.

**What is the expected behavior?**

Missing output files get rebuilt instead of throwing errors.

**Please provide your exact Babel configuration and mention your Linaria, Node, Yarn/npm version and operating system.**

This is with `create-react-app`, Node 8.9.0, Webpack 1.14.0, Babel 6.2.7. This is how I configured Babel:

```
{
    "presets": [
      "react-app",
      [
        "linaria/babel",
        {
          "outDir": "./src/.linaria-cache"
        }
      ]
    ]
  }
```

satya164:
I can successfully reproduce this. This happens due to `babel-loader`'s cache and gets fixed if you delete `node_modules/.cache/babel-loader`.

steadicat:
I actually tried deleting the Babel Loader cache and it didn't seem to work. Deleting the whole `node_modules` directory did work for me though.

satya164:
Hmm that's weird. Maybe you have some other cache too? I tried with ejecting a fresh CRA project and deleting that folder did the trick for me.

zamotany:
Which linaria version are you using?

steadicat:
I can repro with both `0.2.0` and `0.3.1`.

steadicat:
This happens also when you start fresh with a clean repo, and no `outDir` folder. Deleting `node_modules` doesn't seem to help either. This is bad. Is there a workaround? Do we need to touch all the source files or something?

thymikee:
Does building whole project helps? This would restore whole `outDir`.

steadicat:
[...]

PR #186, opened 2017-12-13
Webpack loader

### LINKED ISSUE #73: Investigate how to rebuild parent module, if requirements (child modules) have changed
This issue is connected with #45.

Related https://github.com/kentcdodds/babel-plugin-preval/issues/19

zamotany:
Fixes #73, #169 

Documentation is missing, but I'll make it once the code will be done.

So the JS rule looks like this:
```js
{
  test: /\.js$/,
  use: [
    { loader: 'babel-loader', options: { cacheDirectory: true } },
    { loader: 'linaria/loader' }
  ],
},
```

Then while testing on website:
* changing `theme.js` rebuilds CSS so HMR updates styles
* babel cache is not a problem anymore
* deleting `.linaria-cache` doesn't break anything

### DIFF
--- a/docs/BABEL_PRESET.md
+++ b/docs/BABEL_PRESET.md
@@ -1,5 +1,9 @@
 # `linaria/babel` preset
 
+__If you use Webpack__, we highly recommend to add `linaria/loader` to your JS rule in Webpack config. It supports the same options as the preset. For more information go [here](./BUNDLERS_INTEGRATION.md#loader).
+
+---
+
 In order to have styles in `css` tagged template literals evaluated and extracted you need to add the `linaria/babel` preset to your Babel configuration.
 
 `.babelrc`:
--- a/docs/BABEL_PRESET.md
+++ b/docs/BABEL_PRESET.md
@@ -1,6 +1,6 @@
 # `linaria/babel` preset
 
-__If you use Webpack__, we highly recommend to add `linaria/loader` to your JS rule in Webpack config. It supports the same options as the preset. For more information go [here](./BUNDLERS_INTEGRATION.md#loader).
+__If you use Webpack__, we highly recommend to use `linaria/loader`. It supports the same options as the preset. [See here](./BUNDLERS_INTEGRATION.md#loader) for instructions on configuring the loader.
 
 ---
 
--- a/docs/BUNDLERS_INTEGRATION.md
+++ b/docs/BUNDLERS_INTEGRATION.md
@@ -2,6 +2,54 @@
 
 ## webpack
 
+### Loader
+
+For the best user experience while working with Linaria we recommend to use our custom loader.
+
+Just add it to a JS rule after `babel-loader` in webpack config:
+
+```js
+/* rest of your config */
+module: {
+  /* rest of your module config */
+  rules: [
+    /* rest of your rules */
+    {
+      test: /\.js$/,
+      use: [{ loader: 'babel-loader' }, { loader: 'linaria/loader' }],
+    },
+  ],
+},
+```
+
+It serves the following purposes:
+* no stale CSS
+* omits `babel-loader`'s cache, so you always get CSS generated
+* it triggers CSS generation, when external modules used in `css` tagged template changes
+
+If you use loader, you can remove `linaria/babel` from your Babel config and pass options directly to the loader:
+
+```js
+module: {
+  rules: [
+    {
+      test: /\.js$/,
+      use: [
+        { loader: 'babel-loader' },
+        {
+          loader: 'linaria/loader',
+          options: {
+            single: true,
+            filename: 'styles.css',
+            outDir: 'static'
+          }
+        }
+      ],
+    },
+  ],
+},
+```
+
 ### Development
 
 In order for webpack to pick up extracted CSS files, you need to setup [style-loader](https://github.com/webpack-contrib/style-loader) and [css-loader](https://github.com/webpack-contrib/css-loader).
--- a/docs/BUNDLERS_INTEGRATION.md
+++ b/docs/BUNDLERS_INTEGRATION.md
@@ -4,9 +4,9 @@
 
 ### Loader
 
-For the best user experience while working with Linaria we recommend to use our custom loader.
+For the best developer experience with Linaria, we recommend to use our Webpack loader.
 
-Just add it to a JS rule after `babel-loader` in webpack config:
+In your Webpack config, you'll need to add `linaria/loader` to run Linaria on `.js` files:
 
 ```js
 /* rest of your config */
@@ -21,11 +21,6 @@ module: {
   ],
 },
 ```
-
-It serves the following purposes:
-* no stale CSS
-* omits `babel-loader`'s cache, so you always get CSS generated
-* it triggers CSS generation, when external modules used in `css` tagged template changes
 
 If you use loader, you can remove `linaria/babel` from your Babel config and pass options directly to the loader:
 
--- a/loader.js
+++ b/loader.js
@@ -1,0 +1,1 @@
+module.exports = require('./build/tools/webpack-loader').default;
--- a/src/tools/webpack-loader.js
+++ b/src/tools/webpack-loader.js
@@ -10,30 +10,30 @@ function shouldRunLinaria(source: string) {
   );
 }
 
-function getParserOptions() {
-  // @TODO: find a way to either enable all or enable syntaxes based on babelrc.
-  return {
-    plugins: [
-      'jsx',
-      'flow',
-      'objectRestSpread',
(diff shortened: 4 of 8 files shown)


loader.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./build/tools/webpack-loader').default;
```
thymikee: We want to publish this file, please add it to "files" in package.json

website/webpack.config.js
```diff
@@ -50,7 +50,7 @@ module.exports = (env = { NODE_ENV: 'development' }) => ({
       {
         test: /\.js$/,
         exclude: /node_modules/,
-        use: { loader: 'babel-loader' },
+        use: [{ loader: 'babel-loader' }, { loader: 'linaria/loader' }],
```
satya164: Let's enable cache for `babel-loader` here to make sure we always test against it

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,130 @@
...
+  });
+
+  return {
+    source,
+    map: inputMap,
+    meta,
+  };
+}
+
+function makeLoaderAdapter(fn) {
```
satya164: This abstraction seems unnecessary. Why not call the function directly instead of creating a higher order function to which you pass it as an argument? We don't seem to need it to reuse anywhere.

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,130 @@
...
+      filename,
+      sourceMaps: true,
+      inputSourceMap: map,
+    },
+    new babel.Pipeline()
+  );
+
+  // `transformFromAst` is synchronous in Babel 6, but async in Babel 7 hence
+  // the `transformFromAstSync`.
+  return (babel.transformFromAstSync || babel.transformFromAst)(
```
satya164: Probably better to use the async `babel.transformFromAst`

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,130 @@
...
+
+  return {
+    source,
+    map: inputMap,
+    meta,
+  };
+}
+
+function makeLoaderAdapter(fn) {
+  function loaderAdapter(...args: any[]) {
```
satya164: This code can be hugely simplified by making the result always async:

```js
const callback = this.async();

Promise.resolve()
  .then(() => fn.call(this, ...args))
  .then(
    ({ source, map, meta }) => callback(null, source, map, meta),
    error => callback(error)
  );
```

In this case, it's better to use the second argument of `.then` here for the error callback instead of using `.catch` so that if `callback` throws error, it doesn't get caught and passed to `callback` again. At least that doesn't seem like the intended use of the API.

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,130 @@
...
+function getLinariaParentModules(fs: any, module: any) {
+  const parentModules = [];
+
+  function findLinariaModules(reasons) {
+    reasons.forEach(reason => {
+      if (!reason.module.resource) {
+        return;
+      }
+
+      const source = fs.readFileSync(reason.module.resource).toString();
```
satya164: Better to do `fs.readFileSync(reason.module.resource, 'utf8');` instead of `.toString()`, though not sure if webpack's memory fs supports it.

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,130 @@
...
+function shouldRunLinaria(source: string) {
+  return (
+    (/import .+ from ['"]linaria['"]/g.test(source) ||
+      /require\(['"]linaria['"]\)/g.test(source)) &&
+    /css(\.named)?`/g.test(source)
+  );
+}
+
+function transpile(source: string, map: any, filename: string) {
+  const file = new babel.File(
```
satya164: What happens if the user doesn't have `.babelrc` file at all?

src/tools/webpack-loader.js
zamotany: I got it done, before I found a perfect solution and back them the whole loader was async making it harder to integrate with webpack.

src/tools/webpack-loader.js
zamotany: There is no async version of `transformFromAst` in Babel 6.

src/tools/webpack-loader.js
zamotany: It uses normal `.babelrc`/`package.json`/`.babelrc.js` resulution, so if user don't have a babel config it will just assume that the code is standard ES2015 without any _experimental_ features.

As for inlining config in `babel-loader`, if it's possible to get the options from it through webpack, I do that, otherwise, the loader will accept inlined babel config.

src/tools/webpack-loader.js
zamotany: It supports it, but is there any actual reason to change it?

src/tools/webpack-loader.js
satya164: Yes, I mean always use `babel.transformFromAst`, sync or async. You're handling promises anyway.

loader.js
zamotany: Done

src/tools/webpack-loader.js
```diff
@@ -0,0 +1,125 @@
...
+  // `transformFromAst` is synchronous in Babel 6, but async in Babel 7 hence
+  // the `transformFromAstSync`.
+  return (babel.transformFromAstSync || babel.transformFromAst)(
+    file.parse(source),
+    source,
+    {
+      filename,
+      sourceMaps: true,
+      inputSourceMap: map,
+      presets: [require.resolve('../../babel.js')],
```
satya164: We should be able to pass options to our babel config via the loader options

zamotany: Done

satya164: Hmm, shouldn't it be `presets: [require.resolve('../../babel.js'), loaderOptions]` ?

zamotany: Why? There are other options for babel that you might be interested in. Limiting it only to presets doesn't sound good. What if you want to pass `plugins`?

zamotany: Still, if you want to pass options to linaria preset it would look like this:
```js
{
  loader: 'linaria/loader',
  options: {
    presets: [['linaria/babel', { /* options */ }]]
  }
}
```

satya164: Why would we want to allow passing arbitrary presets and plugins? It's `linaria-loader`. People can use `babel-loader` for other babel stuff.

zamotany: But maybe you don't want all of out plugins to be applied? Only `preval-extract`?

satya164: Not sure what's the usecase, but it should be an item in the options object if it needs to be exposed instead of overriding all plugins/presets.

docs/BABEL_PRESET.md
```diff
@@ -1,5 +1,9 @@
 # `linaria/babel` preset
 
+__If you use Webpack__, we highly recommend to add `linaria/loader` to your JS rule in Webpack config. It supports the same options as the preset. For more information go [here](./BUNDLERS_INTEGRATION.md#loader).
```
satya164: I think the term `JS rule` is confusing. I'd remove it since we have clear instructions for this anyway.

Add `to avoid issues` to the end of the first line.

Change `For more information go [here]` to `[See here] for instructions on configuring the loader.`

docs/BUNDLERS_INTEGRATION.md
```diff
@@ -2,6 +2,54 @@
 
 ## webpack
 
+### Loader
+
+For the best user experience while working with Linaria we recommend to use our custom loader.
```
satya164:
```
For the best developer experience with Linaria, we recommend to use our webpack loader`
```

docs/BUNDLERS_INTEGRATION.md
```diff
@@ -2,6 +2,54 @@
...
+
+Just add it to a JS rule after `babel-loader` in webpack config:
```
satya164:
```
In your webpack config, you'll need to add `linaria/loader` to run Linaria on `.js` files:
```

docs/BUNDLERS_INTEGRATION.md
```diff
@@ -2,6 +2,54 @@
...
+    /* rest of your rules */
+    {
+      test: /\.js$/,
+      use: [{ loader: 'babel-loader' }, { loader: 'linaria/loader' }],
+    },
+  ],
+},
+```
+
+It serves the following purposes:
```
satya164: I think it's unnecessary to mention these since these were bugs/limitations.

PR #191, opened 2018-01-09
Use relative filename when getting styles in stylelint preprocessor

zamotany:
`filename` is absolute (from stylelint) but `rawStyles` has relative paths.
Example:
`filename`: `<CWD>/src/App.js`
`rawStyles`: `{ 'src/App.js': { /* ... */ } }`

### DIFF
--- a/src/tools/stylelint-preprocessor.js
+++ b/src/tools/stylelint-preprocessor.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+import path from 'path';
 import * as babel from 'babel-core';
 import { SourceMapConsumer } from 'source-map';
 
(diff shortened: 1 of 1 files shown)


src/tools/stylelint-preprocessor.js
```diff
@@ -37,11 +37,13 @@ export default function linariaStylelintPreprocessor(/* options */) {
         require.resolve('../sheet.js')
       ).exports.default.rawStyles();
 
-      if (!Object.keys(rawStyles).length || !rawStyles[filename]) {
+      const relativeFilename = filename.replace(`${process.cwd()}/`, '');
```
thymikee: This probably won't work on Windows. How about using `path.relative()` https://nodejs.org/api/path.html#path_path_relative_from_to?

PR #192, opened 2018-01-15
webpack-loader not running on all code

oBusk:
**Summary**

I was running into issues where the linaria `webpack-loader` would not apply on my code written in typescript. Managed to find that typescript outputs tagged templates with a space after the function
```
css `
    background: black;
`;
```

This meant that the regex did not match and code did not run.

**Test plan**

After modifying code, linaria was correctly applied to typescript output.


src/tools/webpack-loader.js
```diff
@@ -7,7 +7,7 @@ function shouldRunLinaria(source: string) {
   return (
     (/import .+ from ['"]linaria['"]/g.test(source) ||
       /require\(['"]linaria['"]\)/g.test(source)) &&
-    /css(\.named)?`/g.test(source)
+    /css(\.named)?\s?`/g.test(source)
```
thymikee: Actually we could adjust it to `\s+` because having more whitespace is also semantically correct

oBusk: Then surely `*` would be appropiate (Zero or more) rather than `+` (one or more)?

thymikee: Oh yea, this is what I meant 😄

oBusk:
@zamotany 👌 Done!

PR #195, opened 2018-01-16
Stylis resolve plugin

### LINKED ISSUE #182: Properly resolve `url(...)` in style strings
## What's broken

Say you have the following file structure:

```sh
├── .linaria-cache
├── src
│   ├── components
│   │   └── App.js
│   └── assets
│       └── bat.png
└── .babelrc
```

Now you write following in `App.js`:

```js
const logo = css`
  background: url(../assets/bat.png);
`;
```

This won't work, because the generated CSS file will be at `.linaria-cache/src/components/App.css`, and when resolving `../assets/bat.png`, webpack will look relative to the the generated file.

## How to fix

We can write a `stylis` plugin which replaces relative paths in `url` with the correct path relative to the generated file.

zamotany:
Fixes #182


src/babel/lib/moduleSystem.js
```diff
@@ -39,6 +39,23 @@ function resolveModuleId(moduleId: string, parent: ?Module): string {
...
+/**
+ * Provide custom module with exports as reference.
+ * Use `@linaria/` prefix to require/import provided module.
+ */
+export function provideModule(
+  moduleId: string,
+  exports: any,
+  isDefault: boolean = false
+) {
+  const moduleInstance = new Module(`@linaria/${moduleId}`, null);
```
satya164: What is `provideModule` used for?

thymikee: It holds a reference to real file so it's possible to get the relative path. We've discussed about @zamotany recently that it's maybe better to change the module name to something like `@linaria-internal-somehash` so it's impossible for someone to mistakenly require it)

PR #207, opened 2018-07-03
fix: only write CSS for file that's being directly processed

### LINKED ISSUE #204: Missing classes in output - depending on filename
Bug

Classes go missing depending on the order in which files are processed. See below for more details.
- Example repo: https://github.com/chrisbateman/linaria-demo
  - `npm install` and `npm run build`
- The issue seems to occur only when using babel-cli and **not** when using babel-loader in a webpack app.

No missing classes regardless of filename

**Please provide your exact Babel configuration and mention your Linaria, Node, Yarn/npm version and operating system.**
- linaria v0.5.0
- node v8.x
- npm v5.x
- macOS

```json
{
  "presets": ["env", "stage-2", ["linaria/babel", {"outDir": "./lib"}]]
}
```

## Input
**a.js**
```js
import {css} from 'linaria';
import {black} from './b.js';

export let red = css`
    color: red;
`;
export let blue = css`
    .${black} & {
        color: blue;
    }
`;
export let green = css`
    color: green;
`;
```

**b.js**
```js
import {css} from 'linaria';

export let black = css`
    color: black;
`;
```

## CSS Output
**a.css**
```css
._black__ad82fc ._blue__1fkrlho{color:blue;}
._green__1fkrlho{color:green;}
```

**b.css**
```css
._black__ad82fc{color:black;}
```

In `a.css` the blue and green classes are both correct - but red class is missing completely. 

However – if your rename `a.js` to `z.js` – the missing red class now shows up in the **other** CSS file (`b.css`):

```css
._red__1y5vt3z{color:red;}
._black__o2cls6{color:black;}
```

Other notes: 
- Moving the red class *below* the blue class will also fix the issue, thought that may not always be doable (if the red class is needed inside the blue class, etc.)
- Exporting a plain string instead of a linaria classname in `b.js` also fixes the issue

satya164:
When evaluating imported modules, they are run through the babel preset, as a result, CSS output for those files are also written.
This could overwrite previously generated CSS with a partially evaluated CSS depending on what's used.
This also resulted in multiple hot reloads in webpack since same files are written multiple times.

Fixes #204

### DIFF
--- a/src/babel/preval-extract/prevalStyles.js
+++ b/src/babel/preval-extract/prevalStyles.js
@@ -62,10 +62,31 @@ export default function(
   ]);
 
   clearLocalModulesFromCache();
+
+  const overrides = process.env.LINARIA_BABEL_PRESET_OVERRIDES || '';
+
+  let parsed;
+
+  try {
+    parsed = JSON.parse(overrides || '{}');
+  } catch (e) {
+    parsed = null;
+  }
+
+  // Avoid writing css from the dependencies
+  // We should only write the CSS for the file the user ran through Babel
+  process.env.LINARIA_BABEL_PRESET_OVERRIDES = JSON.stringify({
+    ...parsed,
+    extract: false,
+  });
+
   const { exports: className } = instantiateModule(
     replacement,
     resolve(state.filename)
   );
+
+  // Restore the plugin options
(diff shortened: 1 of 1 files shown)


satya164:
Not relevant anymore

PR #210, opened 2018-07-12
fix error

dr2009:
**Summary**
this example is error. I copy and run code and find this.

**Test plan**


thymikee:
Thanks for fixing that :)

PR #212, opened 2018-09-02
tests: use Node.js 10

DanielRuf:
**Summary**

**Test plan**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2
 defaults: &defaults
   docker:
-    - image: circleci/node:6.10.3
+    - image: circleci/node:10
   working_directory: ~/linaria
 jobs:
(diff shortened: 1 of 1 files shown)


DanielRuf:
Should we test 6, 8 and 10 in parallel?
See https://github.com/documentationjs/documentation/blob/master/.circleci/config.yml

satya164:
Integration tests haven't been the most reliable. This specific issue will be fixed in https://github.com/callstack/linaria/tree/next, though still need to fix the snapshot, since rendering on the ci is not the same as my machine due to apple changing the default font rendering to antialiased and removing subpixel smoothing

zamotany:
We are investing in new rewrite of Linaria on `next` branch (which use node 10) so I'm going to close that. Nevertheless, thank you for your contribution 😉

PR #213, opened 2018-09-12
[WIP] Refactor to switch to Babel7 and Webpack4

s-panferov:
**Summary**

I *love* this library and I want it to be usable with the modern Babel7/webpack4 stack. I'll do my best to make it work on my own, but I will appreciate any help from the maintainers.


s-panferov:
There is something I don't understand in the current codebase: https://github.com/callstack/linaria/blob/master/src/tools/webpack-loader.js#L108-L124

Why do we need to iterate through the parent modules to do something? Isn't it contradicts with the webpack's loader logic? Why cannot we just work within the boundaries of the current file?

If someone can answer it, I will do a pull-request, because it can be easily fixed at least in webpack part. I see that there is some amount of breakage in Babel7 too.., I need to investigate.

PR #221, opened 2018-09-18
 feat: add stylelint preprocessor

satya164:
**How it works**

The Babel plugin returns the unprocessed data in the `metadata` object. The preprocessor takes this data and constructs a CSS string for stylelint to operate on.

**TODO**

- [x] Verify that we maintain line numbers and column numbers properly in the generated CSS string.
- [x] Sometimes line numbers will be different due to interpolations, we need to update the Babel plugin not to change the number of lines when evaluating an interpolation. It should strip new lines.

### DIFF
--- a/src/babel/extract.js
+++ b/src/babel/extract.js
@@ -10,11 +10,15 @@ const slugify = require('../slugify');
 const { units, unitless } = require('./units');
 
 const hyphenate = s =>
+  // Hyphenate CSS property names from camelCase version from JS string
+  // Special case for `-ms` because in JS it starts with `ms` unlike `Webkit`
   s.replace(/([A-Z])/g, g => `-${g[0].toLowerCase()}`).replace(/^ms-/, '-ms-');
 
 const isPlainObject = o =>
   typeof o === 'object' && o != null && o.constructor.name === 'Object';
 
+// Some tools such as polished.js output JS objects
+// To support them transparently, we convert JS objects to CSS strings
 const toCSS = o =>
   Object.entries(o)
     .filter(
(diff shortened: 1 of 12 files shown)


src/stylelint/preprocessor.js
```diff
@@ -0,0 +1,67 @@
...
+
+        // Append new lines until we get to the start line number
+        let line = cssText.split('\n').length;
+
+        while (line < rule.start.line) {
+          cssText += '\n';
+          line++;
+        }
+
+        cssText += `.${className}`;
```
thymikee: how about single:
```js
cssText += `.${className} {`
```

thymikee:
nvm, I just saw the error on the CI :). Also, looks like `rule.end` is not used anywhere

satya164:
Ready for final review.

src/babel/extract.js
```diff
@@ -250,10 +275,10 @@ module.exports = function extract(
                           // Useful for interpolating components
                           cssText += `.${value.className}`;
                         } else if (isPlainObject(value)) {
-                          cssText += toCSS(value);
+                          cssText += stripLines(toCSS(value));
                         } else {
                           // For anything else, assume it'll be stringified
-                          cssText += value;
+                          cssText += stripLines(value);
```
thymikee: can't we run `stripLines` once and for all at the end?

satya164: @thymikee we don't want to strip lines from all CSS. we want to preserve original line numbers for user-written CSS. just need to strip from interpolated values so that they don't introduce any new lines. this allows us to preserve line numbers with stylelint without complex logic

PR #224, opened 2018-09-21
feat: apply correct specificity when wrapping a styled component with another

satya164:
Previously, if you had a `Title = styled.h1` and `Heading = styled(Title)`, and then have the same CSS properties specified in both, the styles which are later in the stylesheet would take precedence. This is one of the limitations we have (had) compared to styled-components. But if we can evaluate expressions at build time, we can get the parent className and generate a more specific selector, ensuring that the correct styles are applied. This PR aims to implement that.

### DIFF
--- a/src/babel/extract.js
+++ b/src/babel/extract.js
@@ -107,21 +107,18 @@ module.exports = function extract(
 
             let cssText = '';
 
-            Object.keys(state.rules).forEach((className, index) => {
+            Object.keys(state.rules).forEach((selector, index) => {
               mappings.push({
                 generated: {
                   line: index + 1,
                   column: 0,
                 },
-                original: state.rules[className].start,
-                name: className,
+                original: state.rules[selector].start,
+                name: selector,
               });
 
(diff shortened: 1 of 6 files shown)


satya164:
I tested on the website and haven't added tests. If someone wants to help by adding tests, much appreciated :)

src/babel/extract.js
```diff
@@ -166,14 +163,14 @@ module.exports = function extract(
           tag.arguments.length === 1 &&
           tag.callee.name === 'styled'
         ) {
-          styled = { component: tag.arguments[0] };
+          styled = { component: path.get('tag').get('arguments')[0] };
```
thymikee: why this change? is `path.get('tag')` different that `tag` from `path.node`?

satya164: `path.node.tag` will return a `node`, but `path.get('tag')` will return a `path`. We need a `path` to be able to traverse it.

mmakarin:
It's happening to me again. Can you double check it?

PR #226, opened 2018-09-23
docs: expand theming docs to include more approaches

satya164:


docs/THEMING.md
```diff
@@ -1,31 +1,102 @@
...
 
-export default function Header({ title }) {
-  return <h1 className={header}>{title}</h1>;
-}
+## React Context
+
+Another approach is to use React Context to pass down colors, and then use function interpolations with the `styled` tag to use the colors in your component. You could use something like [`@callstack/react-theme-provider`](https://github.com/callstack/react-theme-provider) or write your own HOC.
+
+```js
+const Button = withTheme(styled.button`
```
thymikee: `styled-components` wraps all components with `withTheme` consumer by default, so maybe we should too for better compat? Or at least add such option

docs/THEMING.md
```diff
@@ -1,31 +1,102 @@
...
+  dark: {
+    text: 'white',
+  },
+};
+
+// Create a small helper function to loop over the themes and create CSS rule sets
+const theming = cb =>
+  Object.keys(colors).reduce((acc, name) => Object.assign(acc, {
+    [`.theme-${name} &`]: cb(colors[name]),
+  }), {});
```
thymikee: Hm, `theming` fn could also be applied in the theme provider and propagated through context for easier usage like in example below, or am I missing something?

thymikee:
Btw, can you fix lint-website job?

docs/THEMING.md
satya164: @thymikee it won't work. we need to be able to evaluate the CSS statically. passing via props/context will mean that it will need to be generated during render in runtime. only property values are supported dynamically.

docs/THEMING.md
satya164: not sure about it. linaria is supposed to be zero runtime, so adding this by default increases the runtime (e.g. the callstack theme provider will add around 4 kb). also our `ThemeProvider` is not 100% compatible with `ThemeProvider` from `styled-components`, so it needs separate implementation. migration will still be search/replace without this being built-in.

sshmyg: @satya164 So maybe you can add an ability, to wrap all `styled[...]` components in a row. I mean some setting to make such wrapper in one place, and don't wrap each styled component separately. It's very inconvenient to migrate from `styled-component` for e.g and add wrapper to every component.

Issue #214, opened 2018-09-14
Linaria 1.0 roadmap


satya164:
The next rewrite is at: https://github.com/callstack/linaria/tree/next

### Summary

- Different architecture - Previously the Babel plugin used to write CSS files to the disk, this caused many issues:
  * We have to deal with environment variables to prevent writing files
  * We have to use separate options for testing and SSR to prevent these files from getting written
  * Files keep getting written multiple times when rebuilding in webpack, which in turn triggers multiple rebuilds
  * single: true with babel-cli doesn't work properly as it operates on each file in a separate process which can cause issues like #204
  * Tests are very complex and debugging is harder due to we having to read output from stdout
  Now the Babel plugin writes the CSS as a comment in the file and the webpack loader write the actual files, which avoids all these issues, makes tests significantly simpler.
- Sourcemap support - Users can jump to the JS file where the style was declared from devtools
- Proper dependency tracking - The babel plugin can output the dependency list so that the webpack loader can properly mark the dependencies for the file
- A simplified module system - Previously the module system tried to implement all the methods supported in Node. Now we just support the minimum required API (`module.exports`, `exports`, `require`, `require.resolve`, `__filename` and `__dirname`), which makes the codebase simpler, more maintainable are consistent with webpack
- Restricted sandbox when evaluating code - We prevent access to node's native modules when evaluating because you wouldn't be able to do it with webpack, people should use something like babel-plugin-preval to do those
- React bindings with `styled` helper ala `styled-components` which uses CSS custom properties for dynamic properties

This will be a breaking change and will require using both the Babel plugin and Webpack loader. It'll also address many outstanding issues: #169, #198, #204, #206, #208, #209

### TODO

- [x] Handle dynamic imports when evaluating (should replace them with noop using a Babel plugin)
- [ ] ~~Add back support for `include` in `css` tag (maybe not)~~
- [x] Verify that the `css` and `styled` tags are from linaria when processing them (@zamotany)
- [x] Add `stylelint` preprocessor back (#221)
- [x] Add back `now.json` config for the website and fix deploy (@thymikee)
- [ ] Add missing tests to get the coverage to 100%
- [ ] Fix #217 
- [x] Fix integration tests (@thymikee)
- [x] Review and update documentation
- [x] Update the content of the website (@zamotany)
- [x] Write a CLI to extract the CSS comments into files, useful for libraries (@zamotany)
- [ ] ~~Update https://github.com/thymikee/linaria-jest (@thymikee)~~

[...]

PR #227, opened 2018-09-23
feat: add node API and cli

zamotany:
#214 

Let's decide what options we want to support in CLI and over experience, then I'll add the docs.


src/node.js
```diff
@@ -0,0 +1,44 @@
...
+    resolvedFiles.push(...resolveFile(files));
+  }
+
+  const { babelOptions, sourceMaps = false } = options;
+  return resolvedFiles.reduce((output, filename: string) => {
+    const { code } = transformFileSync(filename, babelOptions);
+    const { css, map } = transform(filename, code, sourceMaps);
+    return {
+      ...output,
+      [filename]: { css, map: map || null },
```
satya164: Also need to set the source content for the source maps so it knows what's in the source JS file. e.g. https://github.com/callstack/linaria/blob/62da50eeb8b46f28caafdebc4d019b30eb9b290d/src/loader.js#L28

bin/linaria.js
```diff
@@ -0,0 +1,59 @@
...
+  console.log(
+    [
+      'Usage: linaria [options] <file1> [...<fileN>]',
+      '',
+      'Arguments',
+      '  <file1> [...<fileN>]  File paths or glob patterns',
+      '',
+      'Options:',
+      '  --help, -h            Show help                    [boolean]',
+      '  --sourceMaps, -m      Generate source maps         [boolean]',
```
satya164: Maybe `-s` is nicer. Other CLIs such as Babel use `-s` to denote source maps.

bin/linaria.js
```diff
@@ -0,0 +1,59 @@
...
+      '  --outDir, -o          Output directory             [string]',
```
satya164: Camelcase isn't common in CLIs, so probably `--source-maps` and `--out-dir` is better instead of `--sourceMaps` and `--outDir`

bin/linaria.js
```diff
@@ -0,0 +1,59 @@
...
+      ? {
+          configFile: config,
+        }
+      : undefined,
+  });
+
+  mkdir.sync(outDir);
+  Object.keys(styles).forEach(filename => {
+    fs.writeFileSync(
+      path.join(outDir, path.basename(filename)),
```
satya164: We would probably also want to preserve the directory structure, otherwise, CSS files may overwrite each other when 2 JS files have the same name

bin/linaria.js
```diff
@@ -0,0 +1,59 @@
...
+    babelConfig: config
+      ? {
+          configFile: config,
+        }
+      : undefined,
+  });
+
+  mkdir.sync(outDir);
+  Object.keys(styles).forEach(filename => {
+    fs.writeFileSync(
```
satya164: We should check if there is any CSS first before writing the file, and if CSS is empty, skip writing

bin/linaria.js
```diff
@@ -0,0 +1,59 @@
...
+      path.join(outDir, path.basename(filename)),
+      styles[filename].css
+    );
+    if (styles[filename].map) {
+      fs.writeFileSync(
+        path.join(outDir, `${path.basename(filename)}.map`),
+        styles[filename].map
```
satya164: Probably should also add a `sourceMapURL` in the CSS file.

zamotany:
@thymikee can you check this? any feedback?

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+      // Add source mapping URL
+      const cssContent = styles[jsFilename].map
+        ? `${
+            styles[jsFilename].css
+          }\n/*# sourceMappingURL=${cssFilename}.map */`
+        : styles[jsFilename].css;
+
+      fs.writeFileSync(cssOutput, cssContent);
+
+      // Add require to original JS file to import extracted CSS.
```
satya164: We don't want to add `require` to the original source files. We should accept a path to the directory where the user has built JS files and insert them there.

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+  );
+} else {
+  const sourceMaps = Boolean(args.s || args.sourceMaps);
+  const outDir = path.resolve(args.o || args.outDir || '');
+  const requireCss = Boolean(args.r || args.requireCss);
+  const config = args.c || args.config;
+
+  const styles = processFile(args._, {
+    sourceMaps,
+    babelConfig: config
```
satya164: Maybe we should set `babelrc: false` if a config is specified?

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+      path.dirname(jsFilename)
+    );
+    const cssFilename = path
+      .basename(jsFilename)
+      .replace(path.extname(jsFilename), '.css');
+    const cssOutput = path.join(outDir, folderStructure, cssFilename);
+
+    // Skip writing if the file already exists and the new CSS code is empty.
+    if (!fs.existsSync(cssOutput) || styles[jsFilename].css) {
+      mkdir.sync(path.dirname(cssOutput));
```
satya164: Will this handle non-existent nested folders? maybe you should use something like https://www.npmjs.com/package/mkdirp

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+    const folderStructure = path.relative(
+      process.cwd(),
+      path.dirname(jsFilename)
+    );
+    const cssFilename = path
+      .basename(jsFilename)
+      .replace(path.extname(jsFilename), '.css');
+    const cssOutput = path.join(outDir, folderStructure, cssFilename);
+
+    // Skip writing if the file already exists and the new CSS code is empty.
```
satya164: Not sure if we should skip writing if CSS file exists. If I change something in the JS and run it again, I'd expect the CSS files to be updated. If we skip it, they'll never be updated unless you delete the existing files first manually.

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+    if (!fs.existsSync(cssOutput) || styles[jsFilename].css) {
```
satya164: Maybe instead of checking `styles[jsFilename].css` here, the node API should avoid adding the file into the result if there's no CSS

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+      mkdir.sync(path.dirname(cssOutput));
+      // Add source mapping URL
+      const cssContent = styles[jsFilename].map
+        ? `${
+            styles[jsFilename].css
+          }\n/*# sourceMappingURL=${cssFilename}.map */`
```
satya164: Think we should add this URL only if we are writing the sourcemaps

src/node.js
```diff
@@ -0,0 +1,48 @@
...
+  const { babelOptions, sourceMaps = false } = options;
+  return resolvedFiles.reduce((output, filename: string) => {
+    const { code } = transformFileSync(filename, babelOptions);
+    const { css, map } = transform(filename, code, sourceMaps);
+    if (map) {
+      map.setSourceContent(filename, fs.readFileSync(filename).toString());
+    }
+    return {
+      ...output,
+      [filename]: { css, map: map || null },
```
satya164: Probably should return `map.toString()` instead of the generator instance

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+      '',
+      'Options:',
+      '  --help, -h            Show help                        [boolean]',
+      '  --source-maps, -s     Generate source maps             [boolean]',
+      '  --require-css, -r     Require CSS in original JS file  [boolean]',
+      '  --out-dir, -o         Output directory                 [string]',
+      '  --config, -c          Path to Babel config file         string]',
+      '',
+      'Example:',
+      '  linaria -m ./file1.js ./file2.js',
```
satya164: This needs to be updated

bin/linaria.js
```diff
@@ -0,0 +1,86 @@
...
+      '',
+      'Arguments',
+      '  <file1> [...<fileN>]  File paths or glob patterns',
+      '',
+      'Options:',
+      '  --help, -h            Show help                        [boolean]',
+      '  --source-maps, -s     Generate source maps             [boolean]',
+      '  --require-css, -r     Require CSS in original JS file  [boolean]',
+      '  --out-dir, -o         Output directory                 [string]',
+      '  --config, -c          Path to Babel config file         string]',
```
satya164: missing brackets?

satya164: also should mention which options are required and throw if not provided. I'd recommend using something like yargs to make it easier instead of writing it manually https://github.com/yargs/yargs

bin/linaria.js
zamotany: this is using mkdirp

bin/linaria.js
zamotany: Please ready carefully: _Skip writing if the file already exists __and the new CSS code is empty___

It was done per your request:
> We should check if there is any CSS first before writing the file, and if CSS is empty, skip writing

bin/linaria.js
zamotany: If source map are disabled `styles[jsFilename].map` will be null so no `sourceMappingURL` will be inserted.

bin/linaria.js
satya164: I didn't mean *if the file already exists*, I said if there is any CSS i.e. if there's any CSS text.

bin/linaria.js
satya164: @zamotany `--out-dir` should be required, no? otherwise would be too easy to pollute working directory with css files by mistake.

zamotany: It would be, I think babel for example doesn't require you to provide out dir. Or maybe we should set some default like `build`? or `css_dist`?

bin/linaria.js
zamotany: Ok got it

bin/linaria.js
satya164: maybe `extracted_css` or something. but maybe better just to require it. I'd want to change it anyways.

satya164:
Would be good to update it per the new `transform` function. Also maybe we should expose `transform` for the public API and the cli for working with files?

satya164:
I rebased the PR against master since there were conflicts.

satya164:
Hey, I fixed the behavior of inserting require statements (it was supposed to insert them in a given directory with compiled JS files, not the source files).

I also replaced `commander` with `yargs` because there was barely any validation of the arguments in `commander`. `yargs` also seems easier to use and the parsed arguments are typed, which is a plus.

Please check and merge when you have time :)

src/cli.js
```diff
@@ -0,0 +1,132 @@
...
+  })
+  .option('source-maps', {
+    alias: 's',
+    type: 'boolean',
+    description: 'Generate source maps for the CSS files',
+    default: false,
+  })
+  .option('source-root', {
+    alias: 'r',
+    // type: 'string',
```
zamotany: Should this be commented out?

satya164: Nope, my mistake

satya164:
```suggestion
    type: 'string',
```

src/cli.js
```diff
@@ -0,0 +1,132 @@
...
+    description: 'Directory contaning the source JS files',
```
satya164:
```suggestion
    description: 'Directory containing the source JS files',
```

PR #229, opened 2018-09-26
docs: rename server rendering docs to critical css

satya164:


README.md
```diff
@@ -76,7 +75,10 @@ module: {
           },
         },
         {
-          loader: 'babel-loader'
+          loader: 'babel-loader',
+          options: {
+            cacheDirectory: false,
```
thymikee: why?

satya164: @thymikee same reason as before. files need rebuilding if something imported changed, so we can't cache it.

maybe we should move babel transpiling to the loader like before, make the babel stuff an implementation detail and not expose it. similar to what @zamotany wanted, but not actually switch away from babel since it's still the best for code transformation.

thymikee: uh oh, imagine using uncached transforms on a 200+ files codebase :|

satya164: @thymikee let's talk about it on discord

PR #231, opened 2018-09-26
fix: check if import is from linaria

satya164:
TODO
- [x] Update tests
- [x] Support paths from the library as well

### DIFF
--- a/src/babel/extract.js
+++ b/src/babel/extract.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-param-reassign */
 /* @flow */
 
-const { relative } = require('path');
+const { relative, dirname } = require('path');
 const generator = require('@babel/generator').default;
 const { isValidElementType } = require('react-is');
 const Module = require('./module');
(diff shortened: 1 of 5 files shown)


src/babel/extract.js
```diff
@@ -71,6 +71,38 @@ const stripLines = (loc, text) => {
   return result;
 };
 
+// Verify if the binding is imported from the specified source
+const imports = (t, scope, name, source) => {
+  const bindings = scope.getAllBindings();
+  const p = bindings[name];
+
+  if (p && t.isImportSpecifier(p) && t.isImportDeclaration(p.parentPath)) {
+    return p.parentPath.node.source.value === source;
```
zamotany: Don't we want to use `require.resolve` for that??

satya164: Yeah, it's in the TODO. I can't use `require.resolve` directly because I want to resolve it relative to the file, so I'll need to use something like https://github.com/sindresorhus/resolve-from

satya164:
pr ready for review

src/babel/extract.js
```diff
@@ -71,6 +71,69 @@ const stripLines = (loc, text) => {
...
+
+  if (!binding) {
+    return false;
+  }
+
+  const p = binding.path;
+
+  const resolveFromFile = id => {
+    /* $FlowFixMe */
+    const M = require('module');
```
zamotany: Should we require it every time?

satya164: I required it inline because there is `Module = require('./module')` in the file scope and I didn't want cause confusion. Requiring again should be ok since it'll be cached after first require, right?

zamotany: It will be, but ut looks less clean. Anyway I'm fine with it.

PR #232, opened 2018-09-30
fix: move linaria's babel plugin to linaria-loader

satya164:
Linaria's babel plugin won't work properly with babel-loader's cache.
It's better to move it to the loader so that we don't have to disable the cache.

### DIFF
--- a/docs/BUNDLERS_INTEGRATION.md
+++ b/docs/BUNDLERS_INTEGRATION.md
@@ -15,16 +15,11 @@ module: {
     {
       test: /\.js$/,
       use: [
+        { loader: 'babel-loader' },
         {
           loader: 'linaria/loader',
           options: {
             sourceMap: process.env.NODE_ENV !== 'production',
-          },
-        },
-        {
-          loader: 'babel-loader'
-          options: {
-            cacheDirectory: false,
           },
         }
       ],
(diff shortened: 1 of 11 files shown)


docs/BUNDLERS_INTEGRATION.md
```diff
@@ -15,25 +15,22 @@ module: {
...
-          },
         }
       ],
     },
   ],
 },
 ```
 
-Make sure that `linaria/loader` is included before `babel-loader`. Setting the `sourceMap` option to `true` will include source maps for the generated CSS so that you can see where source of the class name in devtools. We recommend to enable this only in development mode because the sourcemap is inlined into the CSS files.
+Make sure that `linaria/loader` is included after `babel-loader`. Setting the `sourceMap` option to `true` will include source maps for the generated CSS so that you can see where source of the class name in devtools. We recommend to enable this only in development mode because the sourcemap is inlined into the CSS files.
```
thymikee: unrelated: maybe we should turn this on by default when `NODE_ENV=development`?

website/webpack.config.js
```diff
@@ -29,16 +29,13 @@ module.exports = {
...
+            loader: 'babel-loader',
+            options: require('./babel.config'),
           },
           {
-            loader: 'babel-loader',
+            loader: require.resolve('../src/loader'),
             options: {
-              ...require('./babel.config'),
-              cacheDirectory: false,
+              sourceMap: true,
```
thymikee: check for dev?

PR #237, opened 2018-10-05
fix: use relative paths for slugs to keep it consistent across machines

satya164:


satya164:
test failure seems unrelated

PR #239, opened 2018-10-05
fix: babelrc: false still loads the config file

satya164:


src/transform.js
```diff
@@ -88,8 +88,9 @@ module.exports = function transform(
   const { metadata, code } = babel.transformSync(content, {
     filename,
     presets: [[require.resolve('./babel'), options]],
-    babelrc: false,
     exclude: /node_modules/,
+    babelrc: false,
+    configFile: false,
```
thymikee: we need both?

satya164: @thymikee yup `babelrc: false` for `.babelrc` and `configFile: false` for `babel.config.js`, apparently. Example: https://github.com/facebook/create-react-app/pull/5109/files

PR #240, opened 2018-10-05
fix: fix JS sourcemaps and move stylis out of babel plugin

satya164:
We need to generate sourcemaps when transforming so webpack can show proper sources.

Also moved stylis to the transform helper which would make it not run when evaluating styles and in SSR/test environments, which is unnecessary.

### DIFF
--- a/src/__tests__/__snapshots__/babel.test.js.snap
+++ b/src/__tests__/__snapshots__/babel.test.js.snap
@@ -19,11 +19,14 @@ styled(\\"h1\\")({
 
 exports[`evaluates and inlines expressions in scope 2`] = `
 
-CSS Text: .th6xni0{color:blue;width:33.333333333333336%;}
+CSS:
+
+.th6xni0 {
+  color: blue;
+  width: 33.333333333333336%;
+}
 
 Dependencies: NA
-
-Mappings: [{"generated":{"line":1,"column":0},"original":{"line":3,"column":6},"name":".th6xni0"}]
 
 `;
 
@@ -31,11 +34,11 @@ exports[`handles css template literal in JSX element 1`] = `"<Title class={\\"th
 
 exports[`handles css template literal in JSX element 2`] = `
 
-CSS Text: .th6xni0{font-size:14px;}
+CSS:
+
+.th6xni0 { font-size: 14px; }
 
 Dependencies: NA
-
-Mappings: [{"generated":{"line":1,"column":0},"original":{"line":1,"column":13},"name":".th6xni0"}]
 
 `;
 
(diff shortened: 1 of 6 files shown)


src/__tests__/__snapshots__/babel.test.js.snap
```diff
@@ -147,11 +170,13 @@ styled(\\"h1\\")({
 
 exports[`inlines object styles as CSS string 2`] = `
 
-CSS Text: .th73eyu{position:absolute;top:0;right:0;bottom:0;left:0;opacity:1;min-height:420px;}@media (min-width:200px){.th73eyu{-webkit-opacity:0.8;-moz-opacity:0.8;-ms-opacity:0.8;-o-opacity:0.8;-webkit-border-radius:2px;-moz-border-radius:2px;-ms-border-radius:2px;-o-border-radius:2px;-webkit-transition:400ms;-moz-transition:400ms;-o-transition:400ms;-ms-transition:400ms;}}
+CSS:
 
-Dependencies: NA
+.th73eyu {
+  position: absolute; top: 0; right: 0; bottom: 0; left: 0; opacity: 1; min-height: 420px; @media (min-width: 200px) { -webkit-opacity: 0.8; -moz-opacity: 0.8; -ms-opacity: 0.8; -o-opacity: 0.8; -webkit-border-radius: 2px; -moz-border-radius: 2px; -ms-border-radius: 2px; -o-border-radius: 2px; -webkit-transition: 400ms; -moz-transition: 400ms; -o-transition: 400ms; -ms-transition: 400ms; }
+}
```
thymikee: weird this one is not formatted as the rest?

src/__utils__/linaria-snapshot-serializer.js
```diff
@@ -1,11 +1,14 @@
 module.exports = {
   test: value => value && typeof value.linaria === 'object',
   print: ({ linaria }) => `
-CSS Text: ${linaria.cssText}
+CSS:
+
+${Object.keys(linaria.rules)
+    .map(selector => `${selector} {${linaria.rules[selector].cssText}}`)
+    .join('\n')}
```
thymikee: can't we just run it through prettier? :P

src/__tests__/__snapshots__/babel.test.js.snap
satya164: @thymikee coz it was generated from a JS object while others were handwritten

src/__utils__/linaria-snapshot-serializer.js
satya164: we can, but that wouldn't change anything

satya164:
can you check plz? @zamotany

PR #247, opened 2018-10-12
Make Module.transform configurable from evaluate

giuseppeg:
Replacement of #246

### DIFF
--- a/src/babel/evaluate.js
+++ b/src/babel/evaluate.js
@@ -58,7 +58,8 @@ const resolve = (path, t, requirements) => {
 module.exports = function evaluate(
   path /* : any */,
   t /* : any */,
-  filename /* : string */
+  filename /* : string */,
+  transform /* : ?(text: string) => { code: string } */ = null
 ) {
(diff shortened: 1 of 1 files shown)


src/babel/evaluate.js
```diff
@@ -58,7 +58,8 @@ const resolve = (path, t, requirements) => {
 module.exports = function evaluate(
   path /* : any */,
   t /* : any */,
-  filename /* : string */
+  filename /* : string */,
+  transform /* : ?(text: string) => { code: string } */ = null
```
satya164: Thanks for the PR!

Just a `nit`, but can you extract the `defaultTransform` function to top of this function and use default parameter here? I think it'll be simpler.

giuseppeg: good idea thanks, done :)

giuseppeg: let me know if you prefer `transform` over `transformModule`

giuseppeg:
awesome, thank you guys! :)

PR #254, opened 2018-10-14
fix(webpack): Calculate a hash based on file contents to invalidate WDS

bebraw:
**Summary**

Without this changes to styles won't force WDS to refresh.

**Test plan**

What's the preferred way to test this? I don't think this part is covered by tests yet and likely you should find a way to write the loader without relying on the FS. I know you use to deal with mini-css-extract-plugin but maybe there's a better approach.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -1,3 +1,4 @@
+const crypto = require('crypto');
 const os = require('os');
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


satya164:
Thank you for the PR! Can you check if https://github.com/callstack/linaria/pull/248 fixes the issue?

bebraw:
Sure, I'll give it a go and let you know.

bebraw:
@satya164 What's the preferred way to test? `npm add https://github.com/callstack/linaria/tree/%40satya164/resolve-asset` fails since it's not with standard **package.json** or something like that.

satya164:
```sh
npm install linaria@callstack/linaria#%40satya164%2Fresolve-asset
```

bebraw:
@satya164 Your fix doesn't help for this particular case (forcing page refresh if CSS changes) since the imports still remain the same as far as I can see. That's why I added the hash there.

PR #255, opened 2018-10-14
feat: add basic flow types for css and styled tags

satya164:
screenshot 2018-10-15 at 12 44 48 am


src/react/styled.js
```diff
@@ -11,8 +13,9 @@ function styled(tag) {
       }
     }
 
+    /* $FlowFixMe: Flow doesn't know about forwardRef */
```
thymikee: latest flow should have support for that

src/react/styled.js
```diff
@@ -1,7 +1,9 @@
+/* @flow */
+
 const React = require('react'); // eslint-disable-line import/no-extraneous-dependencies
 
-function styled(tag) {
-  return options => {
+function styled(tag: any) {
+  return (options: any) => {
```
thymikee: no flow-comment syntax here?
btw, options is an object with `vars`, `class` and `name` at least

src/react/styled.js
```diff
@@ -1,7 +1,9 @@
+/* @flow */
+
 const React = require('react'); // eslint-disable-line import/no-extraneous-dependencies
 
-function styled(tag) {
-  return options => {
+function styled(tag: any) {
```
thymikee: should be `React.Element<*>`, since the syntax is `styled(MyComponent)`, no?

zamotany: For `styled(MyComponent)` syntax, `tag` should be `React.ComponentType<*>`. `React.Element<*>` is the type for output from component's render method of returned value from functional component.

thymikee: My bad!

zamotany:
Can we decide if we want to use flow types and strip them or within the comment??

PR #258, opened 2018-10-15
fix: add missing dependency

johanholmerin:
`@babel/plugin-transform-modules-commonjs` is used in `src/babel/evaluate.js` but is missing from dependencies. Causes failure if you don't use it or `@babel/preset-env`, which includes it, in your project.


johanholmerin:
Yes, my bad. Fixed.

PR #259, opened 2018-10-15
fix: force integration test to exit

zamotany:


.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         - run: |
             bash .circleci/install-chrome-deps.sh
             yarn --cwd website build
-            yarn test:integration
+            yarn test:integration --forceExit
```
thymikee: a nice summary of all the commits I've seen here XD

PR #262, opened 2018-10-17
feat: cache babel's AST so we don't have to parse again

satya164:
Not sure how to benchmark this, or if parsing is a bottleneck at all.

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -42,6 +42,8 @@ type PluginOptions = {
 
 const STYLIS_DECLARATION = 1;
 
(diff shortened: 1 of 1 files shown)


src/transform.js
```diff
@@ -42,6 +42,8 @@ type PluginOptions = {
 
 const STYLIS_DECLARATION = 1;
 
+const astCache /* : { [string]: { content: string, ast: Object } } */ = {};
```
thymikee: use Map over Object to store cache for perf in V8

thymikee: We've recently moved `jest-haste-map` for that (20% improvement at FB scale)

PR #263, opened 2018-10-17
feat: add typescript definitions

### LINKED ISSUE #197: TypeScript support

satya164:
I've never used typescript, so not sure if it's correct. But would be a good starting point if someone wants to help with typescript definitions.

fixes #197


index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+declare module "./react" {
+  type InterpolationFunction<T> = (props: T) => string | number;
+
+  type StyledComponent<T> = React.ComponentType<T & { as?: string | React.ComponentType<any> }>;
+
+  type StyledTag<T> = (strings: string[], ...exprs: Array<string | number | object | InterpolationFunction<T>>) => StyledComponent<T>;
+
+  export const styled: {
+    <T>(component: T): StyledTag<React.ElementConfig<T>>,
```
satya164: Here I want to return different types based on the type of `T`, i.e. if `T` is `string`, `StyledTag<{ children?: React.Node, [key: string]: any }>`, otherwise `StyledTag<React.ElementConfig<T>>`

index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+    [tag: string]: StyledTag<{ children?: React.Node, [key: string]: any }>,
```
satya164: Here instead of a generic one, probably will be better if I could do something like:

```ts
div: StyledTag<DivProps>
```

index.d.ts
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+declare module "./index" {
```
Esemesek: It should be `declare module "linaria"`.

index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  export const css: (strings: string[], ...exprs: Array<string | number | object>) => string;
```
Esemesek: Be consistent. Choose between `string[]` and `Array<string>`

satya164: @Esemesek can't man, it's union type

typings/index.d.ts
```diff
@@ -0,0 +1,49 @@
+declare module 'linaria' {
+  export const css: (
+    strings: TemplateStringsArray,
+    ...exprs: Array<string | number | object>
+  ) => string;
+
+  export const cx: (
+    ...classNames: Array<string | false | undefined | null | 0>
```
thymikee: Is 0 something special or can be replaced by number?

satya164: It accepts strings or falsy values. That's why `0` is here, çoz it's falsy.

typings/__fixtures__/index.tsx
```diff
@@ -0,0 +1,30 @@
+/// <reference path="../index.d.ts" />
```
Esemesek: Is this `tripe-slash` thing required on top of each test file?

typings/index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+  type StyledJSXIntrinsics = {
+    [P in keyof JSX.IntrinsicElements]: StyledTag<
+      JSX.IntrinsicElements[P] & { [key: string]: any }
+    >
+  };
+
+  export const styled: StyledJSXIntrinsics & {
+    <T>(component: React.ReactType<T>): StyledTag<T>;
+
+    [key: string]: StyledTag<{
```
Esemesek: Use `Record<string, SomeType>` instead of `{ [key: string]: SomeType }`

typings/index.d.ts
Esemesek: Maybe you should extract to separate type to keep in semantically clean.

typings/__fixtures__/index.tsx
satya164: Yeah, otherwise it can't find the module

typings/index.d.ts
satya164: How would I use the record?

Esemesek: It is more readable.

typings/__fixtures__/index.tsx
resir014: @satya164 You can include the files on `types/tsconfig.json` like so:

```json
{
  "include": [
    "index.d.ts",
    "tests/**/*"
  ]
}
```

Note that you have to put these outside of `compilerOptions`.

types/index.d.ts
```diff
@@ -0,0 +1,47 @@
+// TypeScript Version: 2.8
+
+declare module 'linaria' {
+  function css(
+    strings: TemplateStringsArray,
+    ...exprs: Array<string | number | object>
+  ): string;
+
+  function cx(
+    ...classNames: Array<string | false | undefined | null | 0>
```
resir014: This can probably be extracted into a separate type:

```ts
type FalseyValue = false | undefined | null | 0;

...classNames: (string | FalseyValue)[]
```

typings/__fixtures__/index.tsx
satya164: @resir014 is this necessary? this comment is outdated. with the updated code I don't need this and typescript is checking the code properly

types/index.d.ts
satya164: @resir014 yeah, I did that in a previous commit but then reverted, because it seems that all the type aliases I declare are exported. so the user could do:

```ts
import { cx, Falsy} from 'linaria'
```

Which feels wrong :(

satya164:
Thanks a lot for the review @resir014!

types/index.d.ts
resir014: ah, I see.

PR #283, opened 2018-12-15
fix: support evaluating TypeScript files by default. fixes #27

satya164:


src/babel/module.js
```diff
@@ -58,10 +59,35 @@ class Module {
     this.require.resolve = this.resolve.bind(this);
     this.require.ensure = NOOP;
     this.require.cache = cache;
+
+    // We support following extensions by default
+    this.extensions = ['.json', '.js', '.ts', '.tsx'];
```
thymikee: add `.jsx` as well? I'm cool with not adding that and forcing others to just use `.js` though :D

src/babel/module.js
```diff
@@ -86,15 +112,16 @@ class Module {
       // we would end up in infinite loop with cyclic dependencies
       cache[filename] = m;
 
-      if (/\.(js|json)$/.test(filename)) {
-        // For JS/JSON files, we need to read the file first
+      if (/\.(json|js|tsx?)$/.test(filename)) {
```
thymikee: I think it's safer to generate this regex based on `this.extensions`, but that's not gonna change too often, so feel free to ignore this advice

src/babel/module.js
satya164: death to `.jsx`

satya164: @thymikee I would if I could. But due to syntax ambiguity with deprecated typecast syntax, TypeScript relies on the `.tsx` extension to understand JSX!

PR #287, opened 2019-01-05
feat: allow to ignore modules from processing 

### LINKED ISSUE #286: Painfully slow performance during development
I'm evaluating linaria for a upcoming project and setup a sample repo here: https://github.com/steffenmllr/linaria-ts-storybook

Long story short: If I add linaria to the project the compiling time for 2 sample components increases to a painfully slow startup time (`5s` to `14s`) which I might can live with but the rebuild time goes from instant `200ms` to `5142ms`.

So my question is: Is this a known issue or is my configuration wrong. Also can you point me in the direction where/why this is happening so I can help to improve it :) 

node `v8.14.0` with npm `6.4.1` and linaria `1.0.0-beta.2` see [package.json](https://github.com/steffenmllr/linaria-ts-storybook/blob/a43581271af614a9b20dbb159352cf00afd204f7/package.json#L12-L36)

satya164:
Transpiling a large number of files can cause significant slowdowns, especially visible when transforming files under `/node_modules/`.
This excludes `/node_modules/` by default and adds a config option to customize which files to exclude.

As a next step, we should also cache the files we transpile in-memory to optimize it further.

Fixes #286


docs/BABEL_PRESET.md
```diff
@@ -20,3 +20,4 @@ To use this preset, add `linaria/babel` to your Babel configuration at the end o
 
 * `evaluate: boolean` (default: `true`) - Enabling this will evaluate dynamic expressions in the CSS. You need to enable this if you want to use imported variables in the CSS or interpolate other components. Enabling this also ensures that your styled components wrapping other styled components will have the correct specificity and override styles properly.
 * `displayName: boolean` (default: `false`) - Enabling this will add a display name to generated class names, e.g. `.Title_abcdef` instead of `.abcdef'. It is disabled by default to generate smaller CSS files.
+* `exclude: RegExp` (default: `/node_modules/`) - If you specify a regex here, files matching the regex won't be processed, i.e. the matching files won't be transformed with Babel during evaluation. If you need to transpile certain modules under `/node_modules/`, it's recommended to do it on a module by module basis for faster transforms, e.g. `exclude: /node_modules(?!.*[\/\\](some-module))/`.
```
thymikee: I find this regex to be simpler: `node_modules[\/\\](?!some-module|other-module)`.

Btw, how about using globs? They're generally less confusing than regular expressions and Babel [supports them](https://babeljs.io/docs/en/options#exclude) (to some extent at least).

src/transform.js
```diff
@@ -37,6 +37,7 @@ type Result = {
 type PluginOptions = {
   evaluate?: boolean,
   displayName?: boolean,
+  exclude?: RegExp,
```
thymikee: whole PluginOptions duplicates this: https://github.com/callstack/linaria/pull/287/files#diff-5eee446d80e4d2d68757361db1752ccfL175 maybe worth consolidating.

satya164:
@thymikee updated the PR. had to rename `exclude` to `ignore` to avoid conflict with the `exclude` option in the rollup plugin.

thymikee:
👍

PR #289, opened 2019-01-06
feat: support configuring linaria with a config file

satya164:
The babel preset is used in several ways:
- babel config
- webpack loader
- rollup plugin
- stylelint preprocessor

Having a config file makes it possible to configure it in one place instead of configuring separately for each tool.


src/rollup.js
```diff
@@ -6,10 +6,11 @@ const { createFilter } = require('rollup-pluginutils');
 const transform = require('./transform');
 const slugify = require('./slugify');
 
-type RollupPluginOptions = PluginOptions & {
+type RollupPluginOptions = {
   include?: string | string[],
   exclude?: string | string[],
   sourceMap?: boolean,
+  ...$Shape<PluginOptions>,
```
thymikee: What do you think about making a separate entry for bundlers in the config, so they don't overlap with each other (like with exclude vs ignore)?
We could have global options available for all bundlers, and specific ones if needed.

zamotany:
LGTM, but does that mean the user can add their own babel presets/plugins so that linaria will use them? For example, does this allow to support `babel-plugin-module-resolver`?

src/rollup.js
satya164: @thymikee we could. let's talk about it. probably out of scope for this PR.

satya164:
@zamotany I only support the same options as previous in the config file. though regarding babel transforms, I believe they should be respected during module evaluation if they are in the babel config since we don't disable `babelrc`: https://github.com/callstack/linaria/blob/master/src/babel/evaluate.js#L159

src/rollup.js
thymikee: Yup, not strictly related, just wanted to spark the discussion. e.g.

```js
module.exports = {
  displayName: true,
  bundler: {
    rollup: {
  	  sourceMap: true
    },
    webpack: { 
      // ... 
    }
  }

}
```

satya164: should the configuration file support configuring bundlers? 🤔

thymikee: Maybe not, but it's weird that Rollup config is merged with linaria's :|

satya164: Yeah. We already did that for webpack so it's the same here. Maybe we should remove passing linaria's options from bundler config or make it a separate key there?

thymikee: Mhm, we should only pass the options that make sense to this very bundler.

PR #294, opened 2019-01-14
feat: support custom pre-processor for the CSS

### LINKED ISSUE #142: PostCSS support
feature

**Description**

It would be cool to be able to use PostCSS (included PostCSS syntaxes like SugarSS), is this in the roadmap?

satya164:
This allows the user to specify a custom pre-processor to use instead of stylis as well as disable pre-processing. It essentially makes it possible to use different CSS syntax such as Sass with Linaria.

Fixes #142


src/transform.js
```diff
@@ -83,27 +90,43 @@ module.exports = function transform(code: string, options: Options): Result {
...
-    }
+  let preprocessor;
+
+  if (typeof options.preprocessor === 'function') {
+    // eslint-disable-next-line prefer-destructuring
+    preprocessor = options.preprocessor;
+  } else {
+    switch (options.preprocessor) {
+      case 'none':
+        preprocessor = (selector, text) => `${selector} {${text}}\n`;
```
thymikee: Maybe instead of "none" we should me it "css"?

satya164: It can be confusing because you'd use it if you want to use some other syntax instead of standard CSS no?

thymikee: Yea, just had a second thought on that and `none` makes perfect sense, because we don't "pre-process" stuff then

docs/BUNDLERS_INTEGRATION.md
```diff
@@ -83,8 +83,12 @@ Linaria integrates with your CSS pipeline, so you can always perform additional
...
+- `sourceMap: boolean` (default: `false`): Setting this option to `true` will include source maps for the generated CSS so that you can see where source of the class name in devtools. We recommend to enable this only in development mode because the sourcemap is inlined into the CSS files.
+- `cacheDirectory: string` (default: `'.linaria-cache'`): Path to the directory where the loader will output the intermediate CSS files. You can pass a relative or absolute directory path. Make sure the directory is inside the working directory for things to work properly. **You should add this directory to `.gitignore` so you don't accidentally commit them.**
+- `preprocessor: 'none' | 'stylis' | Function` (default: `'stylis'`): You can override the pre-processor if you want to override how the loader processes the CSS. Note that doing this might prevent the `stylelint` from linting the CSS.
```
ai: Why PostCSS or Sass processor will disable linter if Stylelint supports most of preprocessors?

satya164: @ai I meant when using `stylelint` with our config ('linaria/stylelint'). We set it to Sass syntax, so it will work for Sass/similar syntax like postcss-nested. But it won't work with any other syntax. I think manually configuring it will make it work, just not with the default config. I'll make it clearer.

Pytal:
Is there a way to use a custom pre-processor when using the [Parcel](https://github.com/parcel-bundler/parcel) bundler?

PR #295, opened 2019-01-14
fix(url-rewrite): replace multiple url() in declarations

wardpeet:
**Summary**
It fixes file rewrites for css declarations with multiple src valus
```
css`
@font-face {
  font-family: Roboto;
  src: url(./file.woff2) format('woff2'), url(./file.woff) format('woff');
`
```

**Test plan**
I've added tests to make sure it works, if you run the tests on old code you'll see that it won't do it correctly.


satya164:
Thanks for the PR. I think the regex was just missing the `g` flag. I added it and merged. Thanks for adding the tests!

wardpeet:
sweet, I thought you would lose the segments in your regex, like you do with `.match`. TIL, it isn't the case with replace.

PR #297, opened 2019-01-15
fix: use webpack's resolver to resolve files. fixes #284

### LINKED ISSUE #284: Linaria webpack loader can't resolve webpack alias

bug

webpack config:
```
  resolve: {
    extensions: ['.js', '.jsx'],
    alias: {
      '@': `src/shared`,
    },
  },
```

file with linaria import:
```
import { css } from 'linaria';
import bp from '@/lib/breakpoints';
```

error:
```
Error: Module build failed (from ../node_modules/linaria/loader.js):
 SyntaxError: src/shared/components/TitleWithIcon/TitleWithIcon.styles.js: An error occurred when evaluating the expression: Cannot find module '@/lib/breakpoints'. Make sure you are not using a browser or Node specific API.
```

satya164:
I'm using `this._compilation.options` (private and deprecated API) to get access to webpack's resolve options such as `resolve.alias`, `resolve.extensions` etc.

There seems to be no other way to access webpack's resolver. There is `this.resolve`, but it's asynchronous. There was `this.resolveSync`, but it seems to be removed. `this._compilation` is used by many loaders/plugins such as `mini-cssextract-plugin`, so hope we're safe for a while.

Other options are:
1) read the webpack.config.js manually, but it won't work for programmatic usage. `eslint-import-resolver-webpack` does this.
2) require the user to pass the resolve options in the loader options as well. it'll duplicate webpack's config, but won't break.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -4,9 +4,9 @@ const fs = require('fs');
 const path = require('path');
 const mkdirp = require('mkdirp');
 const normalize = require('normalize-path');
-/* $FlowFixMe */
-const Module = require('module');
 const loaderUtils = require('loader-utils');
+const enhancedResolve = require('enhanced-resolve/lib/node');
+const Module = require('./babel/module');
 const transform = require('./transform');
 
 module.exports = function loader(content: string, inputSourceMap: ?Object) {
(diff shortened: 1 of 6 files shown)


thymikee:
If isn't there already, please create an issue about exposing public API to do so.

src/loader.js
```diff
@@ -47,15 +69,11 @@ module.exports = function loader(content: string, inputSourceMap: ?Object) {
...
-            id: this.resourcePath,
-            filename: this.resourcePath,
-            paths: Module._nodeModulePaths(path.dirname(this.resourcePath)),
-          });
+          const f = resolveSync(path.dirname(this.resourcePath), dep);
 
           this.addDependency(f);
         } catch (e) {
-          // Ignore
+          console.warn(`Failed to add dependency for: ${dep}`);
```
thymikee: maybe we could add error message to the log, so the reason is known?

PR #300, opened 2019-01-16
Multiple stuff

satya164:


src/react/styled.js
```diff
@@ -93,7 +123,7 @@ type StyledTag<T> = (
 
 type StyledJSXIntrinsics = $ObjMap<
   $JSXIntrinsics,
-  <T>({ props: T }) => StyledTag<T>
+  () => StyledTag<{ children?: React$Node, [key: string]: any }>
```
satya164: flow was failing here and due to incomplete dom typings in flow, the same thing would've been the result

src/react/styled.js
```diff
@@ -44,27 +63,38 @@ function styled(tag: React.ComponentType<*> | string) {
...
 
       /* $FlowFixMe */
-      if (typeof tag.className === 'string' && tag !== component) {
+      if (tag.__linaria && tag !== component) {
         // If the underlying tag is a styled component, forward the `as` prop
         // Otherwise the styles from the underlying component will be ignored
         return React.createElement(
           tag,
-          Object.assign(rest, {
+          Object.assign(filteredProps, {
```
thymikee: this mutation is on purpose? would be cool to make it explicit

src/react/styled.js
```diff
@@ -44,27 +63,38 @@ function styled(tag: React.ComponentType<*> | string) {
...
+      ? React.forwardRef(render)
+      : // React.forwardRef won't available on older React versions and in Preact
+        // Fallback to a innerRef prop in that case
+        ({ innerRef, ...rest }) => render(rest, innerRef);
 
     Result.displayName = options.name;
-    Result.className = options.class;
-    Result.extends = tag;
+
+    // This properties will be read by the babel plugin for interpolation
```
thymikee:
```suggestion
    // These properties will be read by the babel plugin for interpolation
```

satya164:
> what about the footprint of adding is-valid-prop?

what can we do about it though?

thymikee:
nothing, just curious what it is right now :D

PR #302, opened 2019-01-24
feat: allow to pass babelOptions, fixes #274

### LINKED ISSUE #274: Allow to configure babel
Feature

Linaria uses [a .babelrc file inside a project](https://github.com/callstack/linaria/blob/c949f19d76cb5027f980c3de864dae0b6f98ea2e/src/transform.js#L67). I believe it works most of the time for small projects. The problem is that in our company babel configuration is centralized, so we don't have a dedicated babel configuration file in every project we have, because it's hard to maintain them independently (we have for about 80 UI applications). We manage babel's configuration inside a shared webpack config file.

Could you please provide a way to just pass a babel configuration inside linaria loader?

P.S. I worked in early alpha versions, I think https://github.com/callstack/linaria/pull/261 changed it.

s-panferov:
**Summary**

This PR adds an ability to pass `babelOptions` to linaria. This is useful in case you have centralized configuration management and don't want to have a `.babelrc` file. Please see the full discussion in #274

### DIFF
--- a/src/babel/evaluate.js
+++ b/src/babel/evaluate.js
@@ -158,18 +158,31 @@ module.exports = function evaluate(
             return { code: text };
           }
 
+          const babelOptions = (options && options.babelOptions) || {};
+
+          if (typeof babelOptions.plugins === 'undefined') {
+            babelOptions.plugins = [];
+          }
+
+          babelOptions.plugins.push(
+            // Include this plugin to avoid extra config when using { module: false } for webpack
+            '@babel/plugin-transform-modules-commonjs',
+            '@babel/plugin-proposal-export-namespace-from',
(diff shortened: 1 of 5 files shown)


src/babel/evaluate.js
```diff
@@ -158,18 +158,31 @@ module.exports = function evaluate(
             return { code: text };
           }
 
+          const babelOptions = (options && options.babelOptions) || {};
+
+          if (typeof babelOptions.plugins === 'undefined') {
```
s-panferov: I'm not sure about the following part, I need help to understand I'm not breaking anything with this.

satya164:
I did some changes to your PR and published a release to npm with the changes.

PR #303, opened 2019-02-07
feat: throw error for invalid interpolations at build-time to avoid mistakes

satya164:
image

image

### DIFF
--- a/src/__tests__/__snapshots__/preval.test.js.snap
+++ b/src/__tests__/__snapshots__/preval.test.js.snap
@@ -362,9 +362,45 @@ Dependencies: NA
 
 exports[`throws codeframe error when evaluation fails 1`] = `
 "<<DIRNAME>>/source.js: An error occurred when evaluating the expression: This will fail. Make sure you are not using a browser or Node specific API.
-[0m [90m 4 | [39m[0m
-[0m [90m 5 | [39m[36mconst[39m [33mTitle[39m [33m=[39m styled[33m.[39mh1[32m\`[39m[0m
-[0m[31m[1m>[22m[39m[90m 6 | [39m[32m  font-size: \${foo()}px;[39m[0m
-[0m [90m   | [39m               [31m[1m^[22m[39m[0m
-[0m [90m 7 | [39m[32m\`[39m[33m;[39m[0m"
+  4 | 
+  5 | const Title = styled.h1\`
+> 6 |   font-size: \${foo()}px;
+    |                ^
+  7 | \`;"
+`;
+
+exports[`throws when interpolation evaluates to NaN 1`] = `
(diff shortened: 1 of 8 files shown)


src/__tests__/__snapshots__/preval.test.js.snap
```diff
@@ -368,3 +368,39 @@ exports[`throws codeframe error when evaluation fails 1`] = `
...
+[0m [90m 7 | [39m[32m\`[39m[33m;[39m[0m"
+`;
+
+exports[`throws when interpolation evaluates to undefined 1`] = `
+"<<DIRNAME>>/source.js: The expression evaluated to 'undefined', which is probably a mistake. If you want it to be inserted into CSS, explicitly cast or transform the value to a string, e.g. - 'String(fontSize)'.
+[0m [90m 4 | [39m[0m
+[0m [90m 5 | [39m[36mconst[39m [33mTitle[39m [33m=[39m styled[33m.[39mh1[32m\`[39m[0m
+[0m[31m[1m>[22m[39m[90m 6 | [39m[32m  font-size: \${fontSize};[39m[0m
+[0m [90m   | [39m               [31m[1m^[22m[39m[0m
+[0m [90m 7 | [39m[32m\`[39m[33m;[39m[0m"
```
thymikee: these snapshots would use `strip-ansi` for readability

src/react/styled.js
```diff
@@ -58,10 +58,31 @@ function styled(tag: React.ComponentType<*> | string) {
...
 
         Object.keys(vars).forEach(name => {
-          const [value, unit = ''] = vars[name];
-          style[`--${name}`] = `${
-            typeof value === 'function' ? value(props) : value
-          }${unit}`;
+          const [result, unit = ''] = vars[name];
+          const value = typeof result === 'function' ? result(props) : result;
+
+          if (process.env.NODE_ENV !== 'production') {
```
thymikee: extract it to a helper invariant

src/babel/extract.js
```diff
@@ -344,12 +344,36 @@ module.exports = function extract(babel: any, options: Options) {
                   : { line: end.line, column: end.column + 1 },
               };
 
+              // If value is null or undefined, it's probably a mistake
+              const throwIfInvalid = (value: any) => {
```
thymikee: maybe move this to the top scope and pass `ex` as an argument to avoid unnecessary recreation of this function

satya164:
> But isn't it a breaking change

Yeah, but the published version is a beta so...

> I'd think about making it a deprecation warning

It's very easy not to notice warnings amongst the webpack output logs though

thymikee:
You're right about beta :)

PR #306, opened 2019-02-08
docs: document how to use global styles

satya164:


website/src/index.js
```diff
@@ -1,9 +1,36 @@
...
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './components/App';
-import './styles/global.css';
+import constants from './styles/constants';
 
 /* $FlowFixMe */
 ReactDOM.render(<App />, document.getElementById('root'));
+
+export const globals = css`
```
thymikee: exported but not imported anywhere?

satya164: @thymikee needed to stop eslint complaining :)

PR #309, opened 2019-02-11
 feat: drop unused CSS from output

satya164:
See [6f6102a](https://github.com/callstack/linaria/pull/309/commits/6f6102a44eee3c93656bd5e0e20ecab0037ba666) for a small diff without indentation changes

### DIFF
--- a/src/babel/extract.js
+++ b/src/babel/extract.js
@@ -268,295 +268,285 @@ module.exports = function extract(babel: any, options: Options) {
           css = t.isIdentifier(tag) && tag.name === 'css';
         }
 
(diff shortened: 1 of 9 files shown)


src/babel/extract.js
```diff
@@ -541,6 +550,10 @@ module.exports = function extract(babel: any, options: Options) {
           path.replaceWith(t.stringLiteral(className));
         }
 
+        if (!isReferenced && !cssText.includes(':global')) {
```
thymikee: Make it explicitly `isReferenced === false `

PR #310, opened 2019-02-11
feat: determine display name from file name

satya164:
Closes #291


src/babel/extract.js
```diff
@@ -272,6 +272,11 @@ module.exports = function extract(babel: any, options: Options) {
           return;
         }
 
+        // Increment the index of the style we're processing
+        // This is used forslug generation to prevent collision
```
zamotany:
```suggestion
        // This is used for slug generation to prevent collision
```

PR #311, opened 2019-02-14
chore: setup husky and commitlint and update CONTRIBUTING.md

satya164:


CONTRIBUTING.md
```diff
@@ -10,41 +8,82 @@ We want this community to be friendly and respectful to each other. Please read
...
+    ```sh
+    yarn watch
+    ```
+
+1. If you've added code that should be tested, add tests.
+
+1. Ensure the test suite passes by running:
+
+    ```sh
+    yarn lint && yarn flow && yarn test
```
thymikee: shouldn't it be under a single command like `test:ci`?

package.json
```diff
@@ -102,6 +105,12 @@
   "resolutions": {
     "**/babel-core": "7.0.0-bridge.0"
   },
+  "husky": {
+    "hooks": {
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
```
thymikee: I have mixed feelings about commit message hook, because all you care about is a PR message, which is gonna be like `satya164/branch-name` if you have more than 1 commit.

And it will make you cry whenever you need to add more commits and prepend them with random `chore:` `fix:` or other that don't really matter.

thymikee:
All in all, my experience with message hooks is that I still need to adjust the PR message by hand and there's no way to enforce that through GitHub

satya164:
> All in all, my experience with message hooks is that I still need to adjust the PR message by hand and there's no way to enforce that through GitHub

Yeah, that's a limitation of GitHub. But some check is still better than no checks.

package.json
satya164: you can just use `git commit -n` if it's a wip commit imo.

thymikee: would be cool to add it to the docs

CONTRIBUTING.md
satya164: we don't run these commands on ci the same way (e.g. we run unit tests and lint in separate jobs)

satya164: Also these should run when committing anyway, so can probably remove this section entirely

PR #312, opened 2019-02-14
Chore: Upgrade src/website to use css-loader 2.x

hinok:
- This commit upgrade css-loader to 2.1.0 to show how webpack's configuration
should look like for new users who may install the newest version of css-loader
that has breaking change regarding css-modules

More info:
https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#breaking-changes

**Summary**

New users may install css-loader directly like 

```bash
npm install css-loader
# or
yarn add css-loader
```

and these commands will install `css-loader 2.1.0`. Since the `2.x` [css-modules are disabled by default](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md#breaking-changes)

> by default css modules are disabled (now modules: false disable all css modules features), you can return old behaviour change this on modules: 'global'

This commit upgrade css-loader to the newest version - `2.1.0` and also set `modules: "global"` in `webpack.config.js` so newcomers won't struggle with the problem "why my global styles don't work".

**Test plan**

Website should look exactly the same as the production version

**Production**

zrzut ekranu 2019-02-14 o 23 38 56
zrzut ekranu 2019-02-14 o 23 39 33
zrzut ekranu 2019-02-14 o 23 39 41

**After changes**

zrzut ekranu 2019-02-14 o 23 39 59
zrzut ekranu 2019-02-14 o 23 40 12
zrzut ekranu 2019-02-14 o 23 40 04

### DIFF
--- a/docs/BASICS.md
+++ b/docs/BASICS.md
@@ -414,7 +414,7 @@ You can do the following to generate unscoped global styles:
 
 ```js
 export const globals = css`
-  :global {
+  :global() {
     html {
       box-sizing: border-box;
     }
--- a/src/__tests__/__snapshots__/babel.test.js.snap
+++ b/src/__tests__/__snapshots__/babel.test.js.snap
@@ -174,7 +174,7 @@ exports[`includes unreferenced styles for :global 2`] = `
 CSS:
 
 .ah6xni0 {
-  :global {
+  :global() {
     .title {
       font-size: 14px;
     }
--- a/src/__tests__/babel.test.js
+++ b/src/__tests__/babel.test.js
@@ -376,7 +376,7 @@ it('includes unreferenced styles for :global', async () => {
       import { styled } from 'linaria/react';
 
       const a = css\`
-        :global {
+        :global() {
           .title {
             font-size: 14px;
           }
--- a/website/package.json
+++ b/website/package.json
@@ -38,7 +38,7 @@
     "babel-plugin-module-resolver": "^3.1.1",
     "cross-env": "^5.2.0",
     "css-hot-loader": "^1.4.2",
-    "css-loader": "^1.0.0",
+    "css-loader": "^2.1.0",
     "del-cli": "^1.1.0",
     "file-loader": "^2.0.0",
     "mini-css-extract-plugin": "^0.4.4",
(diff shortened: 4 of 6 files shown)


hinok:
Not sure why `ci/circleci: lint-and-typecheck` is failing

```bash
Monitor logs will go to /tmp/flow/zShomezScirclecizSlinaria.monitor_log
Lost connection to the flow server (0 retries remaining): /Out of retries, exiting!
error Command failed with exit code 7.
```

satya164:
Thanks for the PR.

I'm a bit confused by what the change means. I thought that css modules were always disabled by default. Didn't know they allowed `global`/`local` in plain CSS.

How does this change affect configuration with Linaria? From the css-loader docs, I see:

> Disable the CSS Modules spec, all CSS Modules features (like `@value`, `:local`, `:global` and `composes`) will not work

Linaria doesn't use any of these features, so this shouldn't matter, no?

Don't worry about Flow. Doesn't seem related to this PR, just general flakyness.

hinok:
Currently `css-loader` is locked on `1.0.0` where `:global` just works.

The problem appears when you install `css-loader` (like I did...) directly and you end with `css-loader 2.1.0`. In `src/website` there is an example how global CSS can be used but it doesn't work anymore with the new `css-loader`.

Code like this:

```css
:global {
  body { color: red };
}
```

should generate CSS as:

```css
body { color: red };
```

Unfortunately with `css-loader 2.x` css-modules are disabled and final CSS looks like that:

```css
:global body { color: red }; /* :( .... */
```

This commit just upgrade `css-loader` and `webpack.config.js` so everything works again with the newest version.

You can easily see that if you upgrade `css-loader` to `2.1.0`, global styles are broken

zrzut ekranu 2019-02-15 o 00 17 30

zrzut ekranu 2019-02-15 o 00 17 49

zrzut ekranu 2019-02-15 o 00 19 00

satya164:
Oops. For some reason I thought it was handled by `stylis` and not `css-loader`.

Can you change `:global` to `:global()`? That way `stylis` should handle it regardless of `css-loader` config. We also need to update https://github.com/callstack/linaria/blob/master/docs/BASICS.md#adding-global-styles

hinok:
Sorry for the confusion, I thought that it's `css-loader` responsibility to handle global styles. I'm not familiar with `stylis`.

PR updated.

satya164:
@all-contributors let's add hinok for docs

PR #322, opened 2019-02-25
Webpack loader - check if this._compilation is undefined

codecorsair:
In my webpack configuration this._compilation was undefined when running this loader. This also checks if `_compilation` is defined before checking if resolve exists. If not, then falls back to the default options as if resolve is not found.

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -37,7 +37,7 @@ module.exports = function loader(content: string, inputSourceMap: ?Object) {
     // There is this.resolve, but it's asynchronous
     // Another option is to read the webpack.config.js, but it won't work for programmatic usage
     // This API is used by many loaders/plugins, so hope we're safe for a while
-    this._compilation.options.resolve
+    (this._compilation && this._compilation.options.resolve)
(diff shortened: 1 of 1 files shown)


src/loader.js
```diff
@@ -37,7 +37,7 @@ module.exports = function loader(content: string, inputSourceMap: ?Object) {
     // There is this.resolve, but it's asynchronous
     // Another option is to read the webpack.config.js, but it won't work for programmatic usage
     // This API is used by many loaders/plugins, so hope we're safe for a while
-    this._compilation.options.resolve
+    (this._compilation && this._compilation.options.resolve)
```
zamotany:
```suggestion
    this._compilation && this._compilation.options.resolve
```

zamotany: This will fix linting

zamotany:
@codecorsair Thank you for your PR 👍 
We are curios as to why this is happening. Can you provide which Webpack version are you using and how are running it (CLI, Node API, ejected CRA, something else)?

PR #325, opened 2019-02-26
fix: remove redundant `hyphenate` from objects interpolation

Anber:
**Summary**

It looks like we do not need to convert keys of nested objects to kebab case because it brakes nested selectors (eg. `'&.camelCaseSelector': {…}` is converted to `&.camel-case-selector {…}`).

**Test plan**

I've added a test for this case.


Anber:
Hooray!
Thank you.

PR #326, opened 2019-02-27
fix: evaluation of functions with their own nested identifiers

Anber:
**Summary**

There is a problem with the evaluation of functions like this
```
    const foo = (k) => {
      const obj = objects[k];
      return obj;
    };
```

It will be transformed to
```
      {
        const foo = k => {
          const obj = objects[k];
          return obj;
        };

        {
          const obj = objects[k];
          {
            module.exports = foo('key');
          }
        }
      }
```

As we can see, `obj` was extracted from the function to the global scope.

**Test plan**

A new test has been added.


satya164:
Thank you!

PR #328, opened 2019-02-27
Properly escaping back-tick in docs

lmammino:
Properly escaping back-tick in docs

Markdown formatting was not proper to display the single back-tick character on GitHub


lmammino:
Thank you for this awesome library ❤️

PR #339, opened 2019-03-06
fix: don't suppress JS parsing errors in stylelint

satya164:
The formatting is a bit messed up due to stylelint's pretty-printing. But all important information is there.

screenshot 2019-03-06 at 5 07 51 pm


thymikee:
Mind sending this to stylelint issue tracker?

satya164:
@thymikee they are doing that intentionally, I think to be able to format the errors the way they do. we ended up stripping away the codeframe text.

PR #336, opened 2019-03-06
feat: add support for config option in the CLI

zamotany:
#334

PR #342, opened 2019-03-07
fix(cli): properly process configFile if provided

Hotell:
**Summary**

while configFile cli was implemented it doesn't work (https://github.com/callstack/linaria/pull/336).

This PR makes it work (processes the actual config file if present).

**Before:**
image

**After:**

image

> Btw using flow was really a PITA from a contributor experience 👀🤔

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -66,6 +66,11 @@ type Options = {
 function processFiles(files: string[], options: Options) {
   let count = 0;
 
+  const configFileOptions = options.configFile
+    ? // $FlowFixMe
+      require(options.configFile)
+    : {};
+
   const resolvedFiles = files.reduce(
     (acc, pattern) => [...acc, ...glob.sync(pattern, { absolute: true })],
     []
@@ -79,9 +84,7 @@ function processFiles(files: string[], options: Options) {
       {
         filename,
         outputFilename,
-        pluginOptions: {
-          configFile: options.configFile,
-        },
+        pluginOptions: configFileOptions,
       }
     );
 
--- a/src/cli.js
+++ b/src/cli.js
@@ -66,11 +66,6 @@ type Options = {
 function processFiles(files: string[], options: Options) {
   let count = 0;
 
-  const configFileOptions = options.configFile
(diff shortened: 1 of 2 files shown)


src/cli.js
```diff
@@ -66,6 +66,11 @@ type Options = {
 function processFiles(files: string[], options: Options) {
   let count = 0;
 
+  const configFileOptions = options.configFile
+    ? // $FlowFixMe
+      require(options.configFile)
```
satya164: The config file can be a lot of different formats. The reason it's not working in your case is that seems we are not respecting the config file in `transform.js`.

So we need to add the following:

```js
const explorer = cosmiconfig('linaria');

...

const { configFile, ...rest } = options.pluginOptions;
  // Load configuration file
  const result =
    configFile !== undefined
      ? explorer.loadSync(configFile)
      : explorer.searchSync();

  const pluginOptions = {
    ...(result ? result.config : null),
    ...rest,
  };
```

Before this line: https://github.com/callstack/linaria/blob/102397437ff90b35af3089f647968bf528e5481f/src/transform.js#L62

satya164:
> Btw using flow was really a PITA from a contributor experience

We want to migrate to typescript, but it'll take time.

Hotell:
changes applied. Please re-review. cheers

satya164:
Refactored and merged in https://github.com/callstack/linaria/commit/96b82849d8050e1268ac5e93997544e3cee69d09

Hotell:
So why I really appreciate this project and your help, this is not very nice behavior to your contributors. Not at all. 

You should explicitly say next time that you'll do it on your own, rather than leaving someone else spent his free time on something which is gonna be tossed away. Not cool at all.

satya164:
@Hotell I had some time so I thought instead of requesting more changes to your PR, I can finish it quicker and do a release. I based my PR on your PR and also gave attribution to you in my commit. The attribution was mistakenly lost on master because it was squashed instead of "rebase and merge" during merging.

I'm sorry this upset you, but it wasn't my intention.

PR #360, opened 2019-03-19
fix: make linaria compatible with babel language feature plugins

dword-design:
**Summary**
Currently the library does not use the babelOptions for AST transformations. While it is commented that this is on purpose, this breaks with some babel plugins, for example @babel/plugin-proposal-nullish-coalescing-operator. When executing this code:
```
const foo = 'test
console.log((foo ?? '') == '')
```
The output is `false`. When executing this code:
```
import { css } from 'linaria'
css`color: red`
const foo = 'test
console.log((foo ?? '') == '')
```
The output is `'test'`. And the reason is that `transformFromAstSync` strips the parentheses if the plugin is not included, resulting in `console.log(foo ?? '' == '')`.

**Solution**
Use babelOptions also for the transformation.

**Test plan**
No tests written since it's a bugfix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/register": "^7.0.0",
     "@emotion/is-prop-valid": "^0.7.3",
+    "babel-merge": "^2.0.1",
     "cosmiconfig": "^5.1.0",
     "dedent": "^0.7.0",
     "enhanced-resolve": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/register": "^7.0.0",
     "@emotion/is-prop-valid": "^0.7.3",
-    "babel-merge": "^2.0.1",
     "cosmiconfig": "^5.1.0",
     "dedent": "^0.7.0",
     "enhanced-resolve": "^4.1.0",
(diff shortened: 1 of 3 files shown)


pbitkowski:
Hi @dword-design, thanks for your contribution! Sorry for long response time, we will be faster in the future!

@satya164 could you take a look? Do we need this?

satya164:
I don't really want to include the transform plugins because they can increase the build-time a lot, and we'll likely run into plugin ordering issues.

IMO changing the meaning of the code when a transform plugin is not included seems like a bug in Babel. Have your opened an issue in the Babel repo for this or is there an issue already open?

dword-design:
@satya164 thanks for your investigation. I haven't checked Babel for this, but it sounds like it's a problem in Babel indeed. I'll do some research when I have time.

dword-design:
An update on this:
The problem seems to sit quite deeply in babel, so I created an issue in the babel repo:
https://github.com/babel/babel/issues/10966

For the moment there is a quickfix for linaria by setting `parserOpts: { createParenthesizedExpressions: true }` for the parser. This adds parenthesizedExpressions while parsing, and these are respected by transformFromAst.

jayu:
@dword-design any update on this?

dword-design:
@jayu Ah yeah I wrote a PR for babel that got approved and it fixes the problem with the nullish coalescing operator.
https://github.com/babel/babel/issues/10966

The other question here is if it is sensible to use the babel parser in contrast to the babel core transformer, since the latter searches for babel configs. I'm not a fan of the babel parser, but I guess that's up to the dev team of linaria :P.

Besides that I moved to a styling approach different from linaria, so I kinda lost sight of this PR. But I'm still open to contributing stuff.

PR #366, opened 2019-03-23
fix: add missing core-js dependency

kryops:
See comments of https://github.com/callstack/linaria/issues/361

**Summary**

This adds a dependency to `core-js@^2.0.0` which is  implicitly required through the Babel build (because of `"useBuiltIns": "usage"` in `.babelrc`).

/lib/loader.js:
```js
require("core-js/modules/es6.regexp.to-string");

require("core-js/modules/es6.regexp.replace");
```

Now that `core-js@3` is out, upgrading a project can break the linaria webpack loader because `core-js`'s import paths changed. Adding the dependency to linaria should take care of keeping a copy of `core-js@2` around.

**Test plan**

This can be tested by linking the modified version of linaria into a project that uses `core-js@3`

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/register": "^7.0.0",
     "@emotion/is-prop-valid": "^0.7.3",
+    "core-js": "^2.0.0",
     "cosmiconfig": "^5.1.0",
     "dedent": "^0.7.0",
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR!

alexfedoseev:
Missing `core-js` breaks my current build after deps update. Appreciate release with this change in.

miuirussia:
When this fix will be released?

MrTibbles:
👂

Issue #299, opened 2019-01-16
Add scope option for libraries


zamotany:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**What is the current behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**

**What is the expected behavior?**

**Please provide your exact Babel configuration and mention your Linaria, Node, Yarn/npm version and operating system.**

jayu:
@zamotany what is a use case for this feature. Do you want to be able to add a prefix to class names generated by linaria?  I would be useful while building component libraries, right?

thymikee:
Yup, avoiding classname conflicts when publishing css from a library was the goal

jayu:
Since code for this feature is already there (#369), but only needs migration to TypeScript, I mark it as good first issue

jayu:
I missed out that we have `classNameSlug` option. So it is possible to have library scoped class name now.

Issue #208, opened 2018-07-03
Feature Request: avoid hash clashes


chrisbateman:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Hashes are generated from the filename, which can lead to clashes in the following cases:
- 2 different versions of the same library where the styles changed between versions (this is my primary use case)
- 2 separate libraries that both use Linaria and have the same variable names in files with the same name (e.g. `styles.js`).

**What is the expected behavior?**
Possible solutions:
- Always generate hashes based on the styles (I did see the discussion in #147 so I'm not sure how feasible you think this)
  - FWIW I believe this is what Emotion does
- Provide an option in `linaria/babel` to customize the classname generation, which puts responsibility on the user to ensure uniqueness

p.s. Sorry to keep opening issues – I think this is the last one! Would be happy to help contribute if I can.

satya164:
You might wanna check my RFC regarding publishing libraries to NPM: https://github.com/callstack/linaria/issues/178

satya164:
We could also default to `[name]@[version]~[filename]` for hashes, and allow customizing it.

zamotany:
I'm going to close this one in favour of #299

Issue #178, opened 2017-11-24
Support and document how to publish modules using Linaria to NPM


satya164:
## Background

Linaria is probably the best option to choose among CSS in JS libraries if you are making a component library which you want to publish on NPM.

- It produces vanilla CSS files, which means anyone can consume them
- It doesn't have a runtime overhead, so regardless what solution the user uses for CSS, this is not going to another CSS in JS lib to the mix

Today I was trying to publish one of my libraries to NPM which uses Linaria but hit a couple of roadblocks. I hope we can address them.

## Blockers

- The inserted imports are absolute paths, which make it impossible to share the stylesheets among different machines
- The class names might not be unique because they are based on developer's folder structure, so if two people named "Dave" have same folder structure, e.g. `Users/Dave/projects/components`, but have libraries with different names, the class names will clash

## Proposal

- Convert imports paths to be relative to the file, pretty-straightforward
- It's impossible to generate globally unique names, but we can have some logic so that the hashes include the package name, for example: instead of `/Users/satya/awesome/App.js`, we use `awesome~/Users/satya/awesome/App.js` to generate the hash, where `awesome` is the name of the package we read from `package.json` if it exists, this assumes package names are globally unique. This won't work properly if different versions same package are nested, but it's impossible to do it.

Currently, I only need the first one, which should be easy to fix. Appreciate your thoughts.

We also need to document the process of publishing libraries using Linaria to NPM.

satya164:
I think `styled-components` could have been a nice solution to this problem, but there are several other CSS in JS libraries too. Which means that if developers start publishing code using CSS in JS libraries to NPM, it's not going to be only styled-components. Developers are going to use their preferred libraries. And for the users using those libraries, they might get multiple CSS in JS libraries included in their bundles, and even multiple copies of the same CSS in JS library are possible when the libraries use different versions.

In addition to this, there's also the question if multiple CSS in JS libs going to conflict with each other. I have no idea, but it doesn't seem far-fetched.

Even if we solve these issues somehow, for people who do server rendering, they have to configure it to generate CSS for each of the CSS in JS libraries used by their dependencies, which is far more worse than configuring their bundler once. And I'm not sure how it will work if the library you depend upon uses a different version of the same CSS in JS library that you use in your app/from a different dependency. When those libraries upgrade/change their dependency on the CSS in JS library, you also have to update your server rendering code.

> If they wanna override it, just add a className and boom, done.

[...]

PR #369, opened 2019-03-31
feat: scope option for libraries

chrisbateman:
**Summary**

see #299, #208, #178 and updates to docs/CONFIGURATION.md.

I wasn't sure what kind of API you had in mind for this feature, so I won't be offended if you want to go in a different direction!

**Test plan**

Added some Jest tests, linked locally and tested scope: false/true/"custom" and verified that different class names were generated for each.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -11,4 +11,7 @@ emoji=true
 
 module.name_mapper='^linaria' -> '<PROJECT_ROOT>/src'
 
+# flow crashes less on circleci with this
+server.max_workers=1
+
 [strict]
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -34,6 +34,10 @@ module.exports = {
 
   If you need to specify custom babel configuration, you can pass them here. These babel options will be used by Linaria when parsing and evaluating modules.
 
+- `classPrefix: string` (default: `""`):
+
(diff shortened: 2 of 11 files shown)


chrisbateman:
Gotcha – that'd work for me! I could make that update later today – any suggestions for the config name? `prefix`? `classPrefix`?

chrisbateman:
@satya164 Let me know if this is more like what you were thinking of!

thymikee:
Just a heads-up: we're working on TS migration in #398 and will let you know once we're done. Would you be so kind and bear with us until it's ready and rebase then?

pbitkowski:
@satya164 can we merge it after TS migration? Could you take a look?

chrisbateman:
@thymikee can do - thanks for the update!

chrisbateman:
@thymikee @satya164 Would there be any value in pointing this PR at the `develop` branch? Or are y'all plannin to merge that to master at some point?

jayu:
Hi, @chrisbateman Do you have the capacity and will to refactor this to the point it would be possible to merge it to current `develop`?

amaranth:
How is this different from setting the existing `classNameSlug` setting to `"MyScope_[hash]"`?

jayu:
Lol I didn't notice that :o Looks like `classNameSlug` PR was introduced after this one was opened. So this PR does not make sense anymore since we have the possibility to add a prefix to the class name. Thanks for that finding @amaranth

Issue #370, opened 2019-04-01
Typescript error when I refer to a component as selector


hemmxwxsoo:
## Environment
- typescript 3.3.3
- linaria 1.3.1

## Description

I got a type error like below when I try to refer to a component.
```
`Argument of type 'FunctionComponent<ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { as?: "symbol" | "object" | "big" | "link" | "small" | "sub" | "sup" | "track" | "progress" | "a" | "abbr" | ... 163 more ... | undefined; }>' is not assignable to parameter of type 'string | number | CSSProperties | ((props: PropsWithChildren<ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { as?: "symbol" | "object" | "big" | "link" | "small" | "sub" | ... 168 more ... | undefined; }>) => string | number)'.
  Type 'FunctionComponent<ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { as?: "symbol" | "object" | "big" | "link" | "small" | "sub" | "sup" | "track" | "progress" | "a" | "abbr" | ... 163 more ... | undefined; }>' is not assignable to type '(props: PropsWithChildren<ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { as?: "symbol" | "object" | "big" | "link" | "small" | "sub" | "sup" | "track" | "progress" | "a" | "abbr" | ... 163 more ... | undefined; }>) => string | number'.
    Type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)> | null' is not assignable to type 'string | number'.
      Type 'null' is not assignable to type 'string | number'.`
```

## Reproducible Demo

```typescript
import { styled } from 'linaria/react';

const a = styled.div`
    width: 1px;
`;

const b = styled.div`
    ${a} b {          // -> error at ${a}
        width: 2px;
    }
`;
```

I think that it has to be fixed on `linaria/react.d.ts`.

from

``` typescript
type StyledTag<T> = <Props = T>(
  strings: TemplateStringsArray,
  ...exprs: Array<
    string | number | CSSProperties | ((props: Props) => string | number)
  >
) => StyledComponent<Props>;
```

to

``` typescript
type StyledTag<T> = <Props = T>(
  strings: TemplateStringsArray,
[...]

PR #372, opened 2019-04-01
fix: fix type error with a component interpolation

hemmxwxsoo:
**Summary**

It allows using component interpolation without typescript error. (#370)

### DIFF
--- a/react.d.ts
+++ b/react.d.ts
@@ -13,7 +13,7 @@ type StyledComponent<T> = React.StatelessComponent<
 type StyledTag<T> = <Props = T>(
   strings: TemplateStringsArray,
   ...exprs: Array<
-    string | number | CSSProperties | ((props: Props) => string | number)
+    string | number | CSSProperties | ((props: Props) => string | number) | StyledComponent<any>
   >
 ) => StyledComponent<Props>;
 


satya164:
Looks like we need some kind of // ts-ignore comment on that line so the CI passes

hemmxwxsoo:
Sorry @satya164  but I think there may be a problem with that code. But I'm not sure.

If I have a code below.

```javascript
export const Form = styled.form`
    color: red;
`;

export const Input = styled.input`
    ${Form}:hover & {
        width: 1px;
    }
`;
```

Then `Input` constant's type become

```typescript
React.FunctionComponent<React.ClassAttributes<HTMLFormElement> & React.FormHTMLAttributes<HTMLFormElement> & {
    as?: "symbol" | "object" | "a" | "abbr" | "address" | "area" | "article" | "aside" | ... 166 more ... | undefined;
}>
```

instead of

```typescript
React.FunctionComponent<React.ClassAttributes<HTMLInputElement> & React.InputHTMLAttributes<HTMLInputElement> & {
    as?: "symbol" | "object" | "a" | "abbr" | "address" | "area" | "article" | "aside" | ... 166 more ... | undefined;
}>
```

I tried to made it but I don't know what is wrong. 😥

react.d.ts
```diff
@@ -13,7 +13,7 @@ type StyledComponent<T> = React.StatelessComponent<
 type StyledTag<T> = <Props = T>(
   strings: TemplateStringsArray,
   ...exprs: Array<
-    string | number | CSSProperties | ((props: Props) => string | number)
+    string | number | CSSProperties | ((props: Props) => string | number) | StyledComponent<any>
```
Esemesek: Can you add a test case that would cover broken ts definitions?

PR #377, opened 2019-04-03
Update react.d.ts

AwesomeObserver:
**Summary**

Improved React types in react.d.ts


pbitkowski:
Hi @AwesomeObserver, thanks for your contribution! 🍺We're currently migrating Linaria to TypeScript, so we will improve type definitions :). If you would like to take a look, check #398 I will make sure that we'll get your changes into consideration.

PR #389, opened 2019-04-14
Combine default HTML elements attributes with user-defined props. (TypeScript, React)

SogoCZE:
**Summary**

When you add interface with custom props to the styled component generic you will lose autocomplete for HTML element attributes. So I combine default HTML elements attributes with user-defined props.

**Test plan**

```typescript
styled.input<{ error : boolean }>
```

before:

before

after:

after

### DIFF
--- a/react.d.ts
+++ b/react.d.ts
@@ -15,7 +15,7 @@ type StyledTag<T> = <Props = T>(
   ...exprs: Array<
     string | number | CSSProperties | ((props: Props) => string | number)
   >
-) => StyledComponent<Props>;
+) => StyledComponent<Props & T>;
 
 type StyledJSXIntrinsics = {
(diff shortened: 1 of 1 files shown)


pbitkowski:
Hi @SogoCZE! Thanks a lot for your input 🎉! @zamotany should we add this to TS migration as well ?

nhooyr:
@pbitkowski what do you mean by "TS migration"?

nhooyr:
This commit fixes another bug where props of the underlying component are not allowed through if you define additional style props.

E.g. if you use styled on a custom component that accepts props,
they should be passed through instead of being limited to the style props

le0nik:
Can this be merged? It's a really small change, but has a big impact on ability to use linaria with Typescript.

PR #396, opened 2019-04-24
Fix ts cli support

### LINKED ISSUE #395: cli: adding a require statement for generated CSS doesn't work with typescript source files
## Environment

macOs 10.14
node 8

## Description

`src` are typescript files

they are compiled via babel to `dist`. CLI will throws an error because it things there is only `.js` in this world :D https://github.com/callstack/linaria/blob/master/src/cli.js#L120

```
linaria --config --out-dir dist/styles --source-root src --insert-css-requires dist src/**/*.tsx src/**/*.ts
```

I'll send a PR, although after last events/attempt I promised myself I would never contribute with and PR to this project again, well hopefully I didn't waste my time again :) ✌️

Hotell:
**Summary**

Closes #395 

**Test plan**

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -109,6 +109,10 @@ function processFiles(files: string[], options: Options) {
           path.relative(options.sourceRoot, filename)
         );
 
+        const normalizedInputFilename = resolveRequireInsertionFilename(
+          inputFilename
+        );
+
         const relativePath = normalize(
           path.relative(path.dirname(inputFilename), outputFilename)
         );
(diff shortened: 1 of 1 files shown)


Hotell:
CI failed because weird flow error, not my fault:
```
 Worker interrupted with signal: sigkill
```

src/cli.js
```diff
@@ -134,6 +138,21 @@ function processFiles(files: string[], options: Options) {
   console.log(`Successfully extracted ${count} CSS files.`);
 }
 
+function resolveRequireInsertionFilename(filename: string) {
+  const TsFileExtension = ['.ts', '.tsx'];
```
pbitkowski: How about `return filename.replace(/(\.tsx)|(\.ts)/g, '.js');`

thymikee: Even simpler:
```js
return filename.replace(/\.tsx?$/, '.js');
```

Hotell: Sure thing. 

I was rather aiming to get more "verbose" code for the future extensions. Like supporting vue files etc.

Hotell:
Hi everyone. was just wandering, if theres's anything else I should do to get this PR merged. Please lemme know

thanks

Alivejke:
It's being a while since this pr was open, can it be merged? I experience same issue using cli and having ts source files.

Issue #217, opened 2018-09-17
[next] 🐞 `collect()` fails on a particular CSS snippet from bootstrap 4


ivan-aksamentov:
**Do you want to request a *feature* or report a *bug*?**
🐞 bug

**What is the current behavior?**
While server rendering, during collection of critical CSS, the following error is being emitted:

```
TypeError: Cannot read property 'match' of undefined
    at main.css:39:3
    at .dev/server/webpack:/node_modules/linaria/lib/server/collect.js:30:1
    at AtRule.each (.dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:110:1)
    at handleAtRule (.dev/server/webpack:/node_modules/linaria/lib/server/collect.js:29:1)
    at .dev/server/webpack:/node_modules/linaria/lib/server/collect.js:53:1
    at .dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:247:1
    at .dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:146:1
    at AtRule.each (.dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:110:1)
    at AtRule.walk (.dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:143:1)
    at .dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:156:1
    at Root.each (.dev/server/webpack:/node_modules/linaria/node_modules/postcss/lib/container.js:110:1)
```

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**

In my crazy setup, linaria receives a particular snippet of CSS, originally from Bootstrap 4 SASS, via `sass-loader` -> `postcss-loader` -> `css-loader` ->  [`extract-css-chunks-webpack-plugin`](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin) -> [`webpack-flush-chunks`](https://github.com/faceyspacey/webpack-flush-chunks), via some hackery. So far so good, but when I proceeded to server-renderer, `collect()` choked on it.
Here is this snippet:

```
@media print {
  h2,
  h3 {
    page-break-after: avoid;
  }

  @page {
    size: a3;
  }

  body {
    min-width: 992px !important;
  }
}
```

It is reproducible without bootstrap/sass/postcss. Just paste the snippet and linaria will gladly kill the server.
Here is the repro:

[...]

PR #398, opened 2019-04-28
chore: migrate more files to TS

zamotany:
* `src/index.js` was not migrated as it's breaking the tests, probably something with resolving linaria import is working as expected
* `src/react/*.js` - same story, even changing the extension causes tests to fail

@satya164 do you have any idea why that is happening? Or how we can solve that?


satya164:
I'll check tonight or in Tuesday

babel.js
```diff
@@ -1,3 +1,3 @@
 /* eslint-disable import/no-unresolved */
 
-module.exports = require('./lib/babel');
+module.exports = require('./lib/babel').default;
```
Anber: This line broke my canary build.

zamotany: Yeah I forgot to refactor this line to use ES modules: https://github.com/callstack/linaria/blob/chore/migrate-to-fs/src/babel/index.ts#L3

pbitkowski:
@zamotany could you include changes from #377 if they make sense?

thymikee:
FYI: https://github.com/callstack/linaria/pull/387

zamotany:
@satya164 @pbitkowski @thymikee 
Finished migration, ported those 2 PRs, everything is passing, website is working just fine, but:
* testing declaration files is not yet done
* stylelint is working - it's not picking up any errors

I probably won't have time to deal with that in the next days, so if someone wants to do it, feel free.

src/loader.ts
```diff
@@ -1,19 +1,21 @@
...
+  this: any,
+  content: string,
+  inputSourceMap: Object | null
+) {
   const {
-    sourceMap,
+    sourceMap = undefined,
     cacheDirectory = '.linaria-cache',
-    preprocessor,
+    preprocessor = undefined,
```
thymikee: this assignment will only happen once `processor` (or `sourceMap` above) is undefined, so this logic is redundant

src/react/styled.ts
```diff
@@ -1,21 +1,19 @@
...
+  name: string;
+  class: string;
   vars?: {
-    [string]: [
-      string | number | ((props: *) => string | number),
-      string | void,
-    ],
-  },
+    [key: string]: [
+      string | number | ((props: any) => string | number),
```
thymikee: prefer unknown?
```suggestion
      string | number | ((props: unknown) => string | number),
```

src/react/styled.ts
```diff
@@ -46,14 +44,14 @@ function styled(tag: React.ComponentType<*> | string) {
       }
     }
 
-    const render = (props, ref) => {
+    const render = (props: any, ref: any) => {
```
thymikee: something like this maybe?
```suggestion
    const render = (props: { [key: string]: unknown }, ref: React.RefObject<unknown>) => {
```
not sure if ref object is different for function refs and createRefs

src/react/styled.ts
```diff
@@ -121,32 +117,42 @@ function styled(tag: React.ComponentType<*> | string) {
...
-|};
+export default ((process.env.NODE_ENV !== 'production'
+  ? new Proxy(styled, {
+      get(o, prop) {
+        prop = typeof prop === 'number' ? prop.toString() : prop;
+        prop = typeof prop === 'symbol' ? prop.toString() : prop;
+        return o(prop);
+      },
+    })
+  : styled) as any) as Styled;
```
thymikee: can't we just cast the Proxy and mark `styled as Styled`?

src/server/collect.ts
```diff
@@ -45,7 +43,7 @@ const collect = (html: string, css: string): CollectResult => {
   };
 
   stylesheet.walkRules(rule => {
-    if (rule.parent.name === 'keyframes') {
+    if ('name' in rule.parent && rule.parent.name === 'keyframes') {
```
thymikee: nice, these are likely bugfixes

src/transform.ts
```diff
@@ -80,22 +80,25 @@ module.exports = function transform(code: string, options: Options): Result {
       sourceFileName: options.filename,
       inputSourceMap: options.inputSourceMap,
     }
-  );
+  )!;
 
-  if (!metadata.linaria) {
+  if (!metadata || !(metadata as any).linaria) {
```
thymikee: I think we could extend the metadata type returned by babel and avoid casting here and there

src/transform.ts
```diff
@@ -117,7 +121,7 @@ module.exports = function transform(code: string, options: Options): Result {
                 // Replace asset path with new path relative to the output CSS
                 path.relative(
                   /* $FlowFixMe */
```
thymikee: remove $FlowFixMe (here and everywhere)

src/loader.ts
zamotany: nope, TS will cry that it has no default initialiser

thymikee: srsly, that's dumb

src/react/styled.ts
zamotany: I have no idea what the correct types are for this file

thymikee: let's leave it for later then, no worries. marking a `// TODO: better types` would do just fine :)

src/react/styled.ts
zamotany: `styled` also needs to be casted to `any` and then to `styled`. Ideally it should not be needed and the types above should be integrated into implementation, but it's time consuming

src/server/collect.ts
zamotany: technically `'name' in rule.parent ` is here to only make TS happy XD

src/transform.ts
zamotany:
```suggestion
```

src/server/collect.ts
thymikee: I remember something about some of the proprties being undefined while we assumed they don't. Found it: https://github.com/callstack/linaria/issues/217. Likely not related, but still better to be safe

src/react/styled.ts
thymikee: yea that's what I'm concerned about. feel free to mark as todo for later

src/react/styled.ts
zamotany: Added a `// TODO` comment for whole `styled` component

Anber:
Let's merge it all and release first alpha! :)

PR #401, opened 2019-05-01
Migrate to release-it v11

webpro:
**Summary**

As I was working on release-it v11, I tried a few projects in the wild and yours was one of them. So why not open a PR if I have the changes already :)

**Test plan**

You can check out my PR and run `npm run release` (without actually releasing) to see whether it does the job for you.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "babel-core": "^7.0.0-bridge.0",
     "codecov": "^3.2.0",
     "commitlint": "^7.5.2",
-    "conventional-changelog-cli": "^2.0.12",
     "del-cli": "^1.1.0",
     "dtslint": "^0.5.3",
     "eslint": "^5.14.1",
@@ -73,7 +72,7 @@
     "prettier": "^1.16.4",
     "react": "^16.8.3",
     "react-test-renderer": "^16.8.3",
-    "release-it": "^10.1.0"
+    "release-it": "^11.0.0"
   },
   "dependencies": {
     "@babel/core": "^7.3.3",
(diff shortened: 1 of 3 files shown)


satya164:
Thank you for the PR :D

I'll give it a try soon.

thymikee:
@webpro seems like the latest is 12.2 now, would you like to give it a try? :)

webpro:
There are [no breaking changes here](https://github.com/release-it/release-it/blob/master/CHANGELOG.md#v12), so can be updated whenever it fits.

package.json
```diff
@@ -83,6 +82,7 @@
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/register": "^7.0.0",
     "@emotion/is-prop-valid": "^0.7.3",
+    "@release-it/conventional-changelog": "^1.0.0",
```
thymikee: this should be moved to devDependencies

webpro: Oops! Also updated the versions.

thymikee:
Looks like yarn is still complaining on outdated lockfile

thymikee:
Thanks! Btw, is there a reliable way to use `release-it` with monorepo? We want to split linaria into separate packages and that would be helpful

webpro:
There's no built-in way to e.g. `npm publish` each package separately. Depending on your needs, this might be a fairly simply script running release-it from each package dir. I've discussed this topic with @danderson00 in https://github.com/release-it/release-it/issues/516 and he started https://github.com/danderson00/release-em. Happy to discuss this topic and/or help out, especially interested in generic use cases.

danderson00:
Hey @thymikee, I suspect our requirements are similar, mostly focused around taking the pain out of managing versions and publishing multiple npm packages. Have a look at release-em, it's new and a bit immature, but I'd be more than happy to work with you to get it to do what you need reliably.

PR #417, opened 2019-05-21
Fix for broken interpolation

Anber:
I've found a bug in the new extractor.

```
export const T1 = styled.h1`background: #111;`;
export const T2 = styled.h2`background: #222;`;
export const T3 = styled.h3`${T2} { background: #333; }`;
```

Extractor finds referenced `T2` and tries to shake the original file in order to evaluate it. As a result, we get code like this:

```
import { styled } from '../react';
const T2 = styled.h2`background: #222;`;

const wrap = fn => {
  try {
    return fn();
  } catch (e) {
    return e;
  }
};

module.exports = [wrap(() => T2)];
```

Looks fine, but in the process of evaluation, T2 will get a class name based on a wrong `state.index` ( `${relative(state.file.opts.root, state.file.opts.filename)}:0` instead of `${relative(state.file.opts.root, state.file.opts.filename)}:1`).

I'm currently looking for a solution. I hope I'll fix it by the end of the week.

### DIFF
--- a/src/__fixtures__/components-library.js
+++ b/src/__fixtures__/components-library.js
@@ -1,0 +1,5 @@
+import { styled } from '../react';
+
+export const T1 = styled.h1`background: #111;`;
+export const T2 = styled.h2`background: #222;`;
+export const T3 = styled.h3`${T2} { background: #333; }`;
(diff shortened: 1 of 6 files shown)


Anber:
I've moved `slug` and `displayName` generation to the earlier stage so that they can be used later in `templateProcessor` even after shaking.

Anber:
This PR is also ready for the review :)

Anber:
@satya164 @zamotany may I do something to help you with this PR?

PR #422, opened 2019-05-26
Changed suggestion for hot style reloading

skywickenden:
Updated the bundlers integration docs to replace suggestion to use  style-loader or css-hot-loader with mini-css-extract-plugin

mini-css-extract-plugin can now handle this internally and css-hot-loader has been discontinued.

This is a doc change. The example code was pulled from a functioning installation.


thymikee:
@satya164 mind having a look?

thymikee:
Thank you for the contribution! 🙂

Issue #427, opened 2019-05-30
Unify API


brandonkal:
## Describe the feature

CSS in JS libraries are maintaining two seperate APIs, namely styled and css. There are good reasons for both, but in react projects, it is an unnecessary complication. I wish to show that with Linaria's approach it is possible to get the best of both (composition, selection, prop interpolation, and automated css variable interpolations)!

## Problem

Having a single consistant and powerful API is important. Using both together creates some edge cases. Exclusively using the current styled API can cause certain components (those that use Hooks) to not render and you quickly end up creating many components with deep nesting. An example of this is an input with a "clear" button. It must be wrapped in a div to contain and position the button but a styled component would cause the ref to refer to the parent div rather than the input as expected.

For this reason and others, Styled Components are also not currently suitable for building more complex components such as those found in UI libraries. It can be done, but it results in deep nesting. Because Linaria's styled runtime size is near zero, Linaria is in a unique position to unify these two approaches.

### Step 1

The first step would be to support a render prop on the styled component, this would solve many issues with building reusable components:

- The exported component can be used as a child selector for overriding styles.
- refs could be passed down without forgoing the benefits that the styled function provides
- default props (e.g. input type="password", button type="button") can be set easily. Styled Components has a seperate attr() function, but simply supplying a custom render function is a more powerful solution.
- Components with hooks can render themselves (fixing that issue) if provided as a render prop.
- shallower render trees. Exported components can include the styles required to function. No more of this:

```jsx
// Styled(PasswordWrapper) This is done so that the exported module can be selected in a parent styled tag
<Input.Password>
    <PasswordWrapper> {/* Root customized component */}
          <ComponentRoot className="InputWrap_i17nyggi">
               <div className="InputWrap_i17nyggi">
                  <input ... />
                  <span onClick={}>Show</span>
               </div>
          </ComponentRoot>
     </PasswordWrapper>
</Input.Password>
```

This is a simple example, but it can get worse, e.g. the component consumer then wraps the component to customize the styles, or more children use styled. It was worse with emotion, where everything with a css prop was automatically wrapped in multiple Context.Consumers.

The styled function provides a lot of utility and adding a render prop would eliminate a case where css must be used instead.

### Step 2

Introduce the **styled.fragment**. The styled API would also change slightly so that it can be used as a replacement for the css function in every case. The babel plugin could optionally would rewrite styled to css where appropriate though this is not required.
This solves the composition issue with styled.

[...]

Issue #234, opened 2018-10-02
Modifier classes


ai:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the expected behavior?**

To have the same modifiers syntax as in [`astroturf`](https://github.com/4Catalyzer/astroturf)

```js
const Button = styled('button')`
  color: black;
  border: 1px solid black;
  background-color: white;

  &.primary {
    color: blue;
    border: 1px solid blue;
  }

  &.color-green {
    color: green;
  }
`;

<Button primary color="green">
```

Compare to `${  }` interpolation it keeps CSS syntax. CSS highlight and autocomplete works.

satya164:
@thymikee @zamotany thoughts?

thymikee:
I think it introduces a level of indirection and complicates stuff. I'm not a fan.

ai:
@thymikee is it OK to simplify a lot of user’s code but increase complexity in a framework. Is how all features ware made?

Why you are not a fan of it?

ai:
> What about prop name clashes if we wrap custom components with styled()?

Can you show a small example? I am thinking that I was worry about it too, but not sure that we are speaking about the same.

satya164:
For example, say:

```js
const Wrapper = styled(MyComponent)`
  &.primary {
    color: blue;
    border: 1px solid blue;
  }
`;

<Wrapper primary />
```

The `primary` prop is also going to be passed to `MyComponent` unless we filter it out. But we don't know supported props, so not possible filter it out.

Wondering something like this is a good idea:

```js
<Wrapper {...mx({ primary: true, color: 'green' })} />
```

Or maybe the user can use this package: https://npmjs.com/package/classnames

I'm not opposed to the idea, but probably need to think a bit more to avoid the edge cases.

thymikee:
Was about to write the same, but Satya 🤖 was quicker!

ai:
@satya164 why we don’t know about supported props?

We have `styled` object from CSS Modules and can look into classes. `astroturf` does the same to detect what properties should be used as style modifiers. https://github.com/4Catalyzer/astroturf/blob/master/src/runtime/styled.js#L23-L25

satya164:
@ai I'm a bit confused about this line:

```js
const hasModifiers = Object.keys(styles).some(
   className => className !== camelName && className !== kebabName,
);
```

Is `styles` the list of `styles` applied to the component and astroturf keeps them in the bundle?

[...]

PR #437, opened 2019-06-07
feats: modifiers and class name optimizations

brandonkal:
**Summary**

This PR adds the following:
- Improves babel to support arrow functions. This means shorter interpolations inside our styles.
- Adds [modifier class](https://github.com/callstack/linaria/issues/234) support. This is what excites me. See my [RFC](https://github.com/callstack/linaria/issues/427) for context and my earlier thought process. Display logic that is a function of props can be used with normal Javascript, though the syntax looks close to CSS attribute selectors.
- Prefix and suffix support
- Generated BEM-like modifiers. No need to name all those modifiers! JS logic is turned into a CSS modifier name, all of this is stripped away after development:
- Tiny class names

This is a step towards my [RFC on unifying the styled and css APIs](https://github.com/callstack/linaria/issues/427). With the modified Linaria, I have converted some of my own UI components from composition with css to modifiers. The result was slightly less code and components that control their own styles.

**Test plan**

I have added Jest tests for new features.

**Note:** I submitted this for feedback. Although I wish for these features to merged soon, there are still some things I will need input on. I haven't received any input on my RFC yet, so I am hoping a PR that implements some of those ideas may be more effective.
This is the feedback I am looking for:
1. How caching would work with the tiny class names feature. Perhaps a seperate cache should be maintained? This may be a non issue as switching from showing displayName at the start of a classname also throws out linaria's cache.
2. Implementation approach for tiny class names feature: Right now, it is performing regex replacents on result.code and result.cssText after calling transform. Because the transform is syncronous, another approach would be to pass the renaming function as an argument of transform().
3. How to add a method to not minimize a classname/variable. It may be desirable to not minimize some classes/variables names in production.

Known Todos:

- [x] The code assumes `props.key` when validating a modifier class. The example below is valid if state is replaced with props. 
```jsx
const Button = state => styled.button`
  &.${[state.primary]} { color: blue }
`
```
- [x] ~~Accept a configurable array of classnames/variables that should not be minifized~~. This is not possible as classnames are generated in the babel transform. So there really isn't a need for this.
- [x] Only shorten class names in production
- [x] Configure suffix and prefix to work after tiny class name step. This way libraries can have classes with short names like `mui_a`

Unrelated areas for improvement found while working on the code:
The __linaria objects for babel interpolation are not used by styled at runtime. They should be removed from production output.

An example modifier class:
```jsx
export default function Index() {
  return (
    <Page primary>
      <Header />
      <Hero />
    </Page>
  );
}

const Page = props => styled.div`
  background: linear-gradient(to bottom right, #b24592, #f15f79);
  color: #fff;
  min-height: 100vh;
  text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.08);
  &.${[props.primary]} {
    color: #b2ffe6;
  }
`;
```


brandonkal:
This is ready for review. I now consider these features implemented and complete! Thanks.

pbitkowski:
Hi, we're happy that you send your PR, but you have implemented the RFC but we haven't had chance to read and comment it. Let's discuss it further on the RFC issue and then decide what to do about it.

brandonkal:
Yes, let's do that! Thanks.

brandonkal:
Note for those who care: New changes/fixes are being pushed to a seperate branch and are published [@brandonkal/linaria](https://www.npmjs.com/package/@brandonkal/linaria).

Here is a summary of the improvements.

1. Much smaller styled runtime! No dependency on @emotion/is-prop-valid.
2. Simple and predictable prop filtering. Props that end in "$" are not forwarded to the DOM.
3. Modifiers! All style logic can be contained inside your CSS.
4. Improved babel support for arrow functions. This means shorter interpolations inside our styles.
5. Generated BEM-like modifiers. No need to name all those modifiers! Javascript logic is transformed into a CSS modifier name, all of this is stripped away after development:
6. Tiny class names in production with optional prefix and suffix support
7. Uses postCSS rather than stylis. This allows for a much deeper integration with your existing CSS pipeline.
8. High resolution source maps. Even nested selectors and modifiers point to the proper line rather than the start of the template literal.
9. Vastly improved Typescript types and support.

I understand people are busy, but I just wanted to point out that there are other useful improvements in here (e.g. Typescript fixes) that could be merged.

I know this is a big PR, so perhaps the list above is helpful for review. I would be happy to break this PR into several smaller ones. But it has been 53 days since trying to start a discussion here. I see that there is some recent activity on this repo, so I just wanted to post this note to ensure this is not forgotten and work is not duplicated.

brandonkal:
Requesting a response.

Issue #445, opened 2019-07-03
Add minimal runtime for dynamic props with IE 11


nhooyr:
**Do you want to request a *feature* or report a *bug*?**

Feature.

**What is the current behavior?**

dynamic styled props do not work with IE 11.

**What is the wanted behavior?**

They should. It'd be nice if Linaria had a very minimal runtime to modify the stylesheet and dynamically adjust css variables based on passed in props such that they work in IE 11.

I believe this is possible but I could be wrong.

Related: https://github.com/jhildenbiddle/css-vars-ponyfill/issues/5

The lack of IE 11 support is a production blocker for Linaria for many projects.

See https://github.com/facebook/create-react-app/issues/5224
See https://www.reddit.com/r/reactjs/comments/aojnyd/announcing_linaria_10_zero_runtime_cssinjs_library/eg1xsvm?utm_source=share&utm_medium=web2x

klimashkin:
I don't think it's worth it. IE11 is going away, even big enterprise companies that I work with are dropping it this year.

nhooyr:
@klimashkin See #446, the implementation is tiny.

nhooyr:
Going to close as it has been a while and I'm not sure if this is relevant anymore.

Jayphen:
I wonder why people don't use data attributes to implement custom properties?

```css`
  .button {
    color: red;
    &[data-blah] {
      color: blue
    }
}


function Thing({condition}) {
 return (<div data-blah={condition}>hi</div>)
}
```

Is there some downside to this approach that I am missing?

frenzzy:
@Jayphen it does not work if you need to use dynamic style from API or database.
For example a color customizable by user.

frenzzy:
There are a relatively new polyfill exist: https://github.com/nuxodin/ie11CustomProperties
The only requirement for it to use special `ie-style` attribute in addition to `style`:
> There is no way to get the raw content of style-attributes in IE11. Use `<div style="--color:blue" ie-style="--color:blue">` for it

Maybe add a flag to linaria (like useIEStyleAttrubute) which will use both attributes will be much simpler solution here?

PR #446, opened 2019-07-03
Minimal IE Runtime

nhooyr:
See #445

**Summary**

Adds a very minimal runtime for IE. Not the most robust but should work well for our purposes.

See #445

**Test plan**

Not really sure yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "@callstack/eslint-config": "^3.0.2",
     "@commitlint/config-conventional": "^7.5.0",
     "@release-it/conventional-changelog": "^1.0.2",
-    "@types/react": "^16.8.4",
     "all-contributors-cli": "^6.1.1",
     "babel-core": "^7.0.0-bridge.0",
     "codecov": "^3.2.0",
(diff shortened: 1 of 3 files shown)


nhooyr:
The way it work is that when a css variable needs to be modified, instead of just adding an inline style, the IE code will find the rule in the stylesheet that uses the css variable on the class, replace the variable with the value and add this new rule to a dynamically injected stylesheet.

nhooyr:
Not sure if the rule cache here is necessary. For the rule on every dynamic CSS variable, is the selector just the class?

pbitkowski:
Thanks for your contribution! Is this possible to extract this logic to another library, to make this opt-in feature? I know it's tiny, but we don't to do it by default to keep our core small and happy.

dreyks:
@pbitkowski how do you feel about making it a separate library indeed but scope it under callstack? that way the core could still be kept small but the library won't get lost in the future?

nhooyr:
Don't have the time to work on this anymore. Someone else feel free to continue.

Issue #450, opened 2019-07-06
Accept :global("") for global css


nhooyr:
With the current :global() syntax, [WebStorm](https://www.jetbrains.com/webstorm/) complains that it expects a term in between the parenthesis, i.e it cannot be empty. This makes sense as I don't think there is any "real" css pseudo selector that doesn't take an argument.

In order to satisfy WebStorm, it'd be nice if I could instead specify `:global("")` for my global styles.

nhooyr:
Interestingly, `:global(html)` already works. I think this is a stylis feature and not Linaria so closing:  https://github.com/thysultan/stylis.js/blob/master/README.md

PR #451, opened 2019-07-06
Use correct :global syntax in docs

nhooyr:
See #450

### DIFF
--- a/docs/BASICS.md
+++ b/docs/BASICS.md
@@ -414,7 +414,7 @@ You can do the following to generate unscoped global styles:
 
 ```js
 export const globals = css`
-  :global() {
+  :global(*) {
     html {
       box-sizing: border-box;
     }


satya164:
I don't think we want this in the examples. The `*` selector is slow and nesting it like this will also increase the specificity of the global styles.

> This makes sense as I don't think there is any "real" css pseudo selector that doesn't take an argument

I don't think there's a pseudo-selector that takes an argument either. So it's just about the syntax Webstorm supports.

nhooyr:
> I don't think we want this in the examples. The * selector is slow and nesting it like this will also increase the specificity of the global styles.

Any way we can remove the `()`? i.e just `:global`?

> I don't think there's a pseudo-selector that takes an argument either. So it's just about the syntax Webstorm supports.

https://developer.mozilla.org/en-US/docs/Web/CSS/:not

```css
/* Selects any element that is NOT a paragraph */
:not(p) {
  color: blue;
}
```

nhooyr:
Someone else feel free to continue.

PR #453, opened 2019-07-06
Tell stylelint to ignore :global css selector by default

### LINKED ISSUE #443: stylelint complains about :global selector
```
// https://github.com/callstack/linaria/blob/60962090410ea3878f3c190817c1667a124d7df9/docs/BASICS.md#adding-global-styles
export const globalCSS = css`
  :global() {
    @import-sanitize;
  }
`
```

Running stylelint on the above code will give:

```
src/App.tsx
 9:3  ✖  Unexpected unknown pseudo-class selector ":global"   selector-pseudo-class-no-unknown
```

My stylelint config is:

```
extends:
  - stylelint-config-recommended
  - linaria/stylelint-config
```

nhooyr:
Closes #443

### DIFF
--- a/src/stylelint/config.js
+++ b/src/stylelint/config.js
@@ -8,5 +8,11 @@ module.exports = {
     'no-empty-source': null,
     'no-extra-semicolons': null,
     'no-missing-end-of-source-newline': null,
+    'selector-pseudo-class-no-unknown': [
+      null,
+      {
+        ignorePseudoClasses: ['/^global$/'],
+      },
+    ],
   },
 };


nhooyr:
@satya164 What do you think of also enabling `no-descending-specificity` by default? I don't think it makes sense for Linaria as you might have a couple components in a file where the selectors happen to overlap. stylelint shouldn't decide which component appears first in the code.

nhooyr:
Actually nvm, I think in general if `no-descending-specificity` is triggered, something funky is going on with your styles.

nhooyr:
Someone else feel free to continue.

Issue #448, opened 2019-07-04
Consider simplifying new styles evaluator


pbitkowski:
After some investigation the burden of maintaining new extractor is too high. Having to list and manage explicitly all of the syntaxes it's not feasible for us. 

There are still open questions:
1. What additional use cases the new extractor handles than the previous one? 
2. Can we use new extractor logic and get rid of the AST specifics (list of specific language syntaxes / identifiers) ? 

CC: @Anber @thymikee @satya164 @zamotany

Anber:
At least it works with any code base, not only with specially prepared. The current one just fails on our project with out of memory exception and returns a lot of `undefined` instead of classes.

pbitkowski:
Could you provide repro?

Anber:
An example with `undefined` was in original PR:
```
const object = { fontSize: 12 };
object.fontWeight = 'bold';
export default styled.h1`
  ${object};
`;
```

Out of memory happens on our project with 500+ jsx files and I'm afraid, I can not just make a sample repo. It happens because of the current extractor tries to import unnecessary dependencies that also has dependencies…

thymikee:
The main concerns we have is about the maintenance, because JS and other compile-to-js languages (e.g. TS, Flow) tend to add new syntaxes once in a while. We unfortunately don't have unlimited bandwidth to contribute to this project. I hope that's understandable. 

Do you think you could help us with that and maintain this part of the codebase? We'd really appreciate that. We admire the work that you did. It also seems to improve some performance issues we had with previous approach (likely fixed with caching, but still). 

I'd like us to think about further simplifying this implementation, e.g. by
- reducing the surface area 
- figuring out if we can subtract some of the code with existing solutions (like code elimination)

Anber:
> Do you think you could help us with that and maintain this part of the codebase?

Definitely yes, because we have two big projects that we have already migrated to linaria.

> figuring out if we can subtract some of the code with existing solutions
We can If there is an existing solution. It would be the best option.

Supporting for other languages is not as hard as it sounds. For example, for TS I have just added ≈30 lines of code. The main problem is to find broken samples of code (eg. enums for TS).

thymikee:
[...]

PR #462, opened 2019-07-23
WIP: Optimization and simplification of styles evaluator

Anber:
**Summary**

In this PR I'll try to simplify the styles extractor and evaluator (https://github.com/callstack/linaria/issues/448).

**Test plan**

- [x] 100% test coverage


thymikee:
I like the direction where it's going 👍

brandonkal:
This looks like a good direction. I'd like to make a suggestion to improve caching:

The evalCache looks mostly redundant as module evaluations are already cached for Module.require() calls. You will likely see much better build-time wins by caching the work of the babel transform and shaker, persisting the transformed and shaken code to disk.

This will make subsequent builds much faster. To this end, it would be preferrable to not remove exports and instead place the preval exports on a new key: `module.exports.__linariaPreval = [...]`

This way, it is simply a matter of checking if the file source file has been updated since it was last transformed, much like @babel/register does.

This also fixes the issue of the shaker currently only running on babel entrypoints.

brandonkal:
Also, rather than re-implementing a logger here, using the [debug](https://www.npmjs.com/package/debug) module would be preferred and more consistant with other babel plugins.

Anber:
Hi @brandonkal,
thank you for your review!

We increased the number of linaria components in our projects and faced some performance issues, so `evalCache` here is a temporary solution which currently reduces build-time in our projects by the factor of 4. In the near future, it will be replaced by a better solution. The idea with `__linariaPreva` looks interesting. I'll try to implement it.
The main problem right now is supporting different dialects of JS. The simplest solution is extracting this support into plugins (something like src/babel/evaluate/langs/es2015.ts), but I still have a feeling that we can just transpile whatever we meet to ES5 before shaking and evaluation. I have been playing with this idea for the last month, but there are still some pitfalls.

And yes, it's time for improvement debug layer :)

brandonkal:
To transpile before shaking could be worth it. Currently linaria expects to be the first transform (mostly because of preval support). Getting linaria to function as the last transform would be the first step ideally.

I would say rather than targetting ES5, ES6 should be the target (essentially only syntax Node understands). This way the user's config can be used. Here you can see I use the user babel config but override babel-preset-env to target `node: current`: [evaluate.ts#L143-L164](https://github.com/brandonkal/linaria/blob/b898eb5a21a2cfb4a26325663b19111ea57a4dbe/src/babel/evaluate/evaluate.ts#L143-L164).

I've pushed my current work in progress on the `__linariaPreval` idea to that same repository if you are interested in looking at that. Right now I am working on getting exports to not be removed.

You can also see that I've been using the debug module and improved caching.

Anber:
The major change here is extracting evaluators to plugins. Also, I've returned the old extractor (but a bit modified) as a default evaluator. Different files in a project can be processed by different extractors (`rules` option has been introduced).

The new evaluator (`shaker`) now pre-transpiles files to the basic JS and only then shakes them. That means that we don't need to support JS-dialects implicitly. It perfectly supports pure functions but has problems with side-effects. Solution for these problems will cause significant code complexity and that's why I've decided to move this logic to the plugin.

And the last change here is processing dependencies through the evaluators. It reduces the build-time for cases when a component or identifier from an external file is referenced in styles.

There is still some work with tests, documentation and some dead code but I think it's ready for review.

brandonkal:
@Anber you should have a look at my fork again. It takes a different approach by not relying on a shaker. It has HMR support and caches transforms for evaluation. I am interested in seeing a comparison of initial and subsequent build times for the repos you work with.

Anber:
@brandonkal Unfortunately, it's impossible to evaluate styles in a big project without shaking. Your approach also throws `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` on our codebase.
It's possible if you have specially prepared components without many dependencies, but in the real world, it can cause unpredictable and tricky errors. For example, we use i18n in our components and a special prop-type for translatable props (like `message: anyValidMessage.isRequired`). That means, that attempt of evaluation will also cause evaluation of a module which declares `anyValidMessage` (despite the fact, that it is unnecessary for styles), and that module also has some dependencies, one of them is react-intl, which cannot be evaluated in nodejs environment at all (before version react-intl@3.0). And it is just one case of dozens we faced during migration to Linaria.

However, we can try to merge our codebases and you can move your evaluation strategy to a plugin.

brandonkal:
@Anber Thank you for looking into that. Would it be possible to extract your shaker into its own babel plugin? I can see it being useful outside of Linaria. I can see how not shaking could be an issue if prop-types are not stripped by babel or modules that don't function in nodejs are used.

Regarding your results, I would be curious to know if you have any circular dependencies and where that OOM may be coming from.

I would like to try merging codebases.
My main goal with the different evaluation strategy was to only transform and evaluate each file once and then use aggressive caching. That sped things up significantly for my ~70 component component library I am developing with Linaria. Having a shaker in that process would be beneficial for edge cases. I ripped it out in the interim as it is not something I have the bandwidth to maintain on my own. If it was a standalone babel plugin, merging would be easier.

Anber:
@brandonkal sorry for a long time silence.
The extraction of the shaker doesn't make much sense because the shaker uses some Linaria-related optimizations such as removing JSX or prop types.

OOM definitely is caused by circular dependencies or something similar, but I don't even sure that it's happened in my code or somewhere in the dependencies.

I've finished with this PR. There are a lot of touched and moved files, but still no breaking changes and all old tests a still alive.

Now I have to actualize the develop branch and then it will be ready for merge. I hope, @zamotany, @thymikee or @satya164 will come and tell us their decision :)

Anber:
Hm… I’ll split this PR to three different logical part. It will be more clear.

Anber:
It has been split to #497, #498 and #499

PR #482, opened 2019-09-06
feat: add classNameMode option for displayName render

chrisabrams:
**Summary**

At GoDaddy we need the ability to generate a set of .css files which have consistent class names. Having the hash in/as the class name will not work for us as we have teams that only use our CSS (they don't have any React components).

This PR would enable GoDaddy to use Linaria as it provides the ability to generate a CSS class name which is consistent to the `displayName`. In using this new option, the end user takes responsibility for any class name collisions.

## How to configure

```
module.exports = {
  displayName: true,
  classNameMode: 'simple',
};

```

## Example

A file called `header.jsx`:

```
import colors from './colors'
import { css } from 'linaria';
import { modularScale, hiDPI } from 'polished';
import fonts from './fonts';

// Write your styles in `css` tag
const header = css`
  color: ${colors.heading};
  text-transform: uppercase;
  font-family: ${fonts.heading};
  font-size: ${modularScale(2)};

  ${hiDPI(1.5)} {
    font-size: ${modularScale(2.5)};
  }
`;

export default header
```

### Current generated class name

    .header_h1h1jlnn

### Proposed generated class name when using "simple" as the `classNameMode`

    .header

**Test plan**

I ran `yarn flow`, `yarn typescript`, `yarn lint`, and `yarn test`. The docs say there is a `test:integration` option but I did not see that in the `package.json`.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -226,6 +226,17 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "chrisabrams",
+      "name": "Chris Abrams",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/527740?s=460&v=4",
+      "profile": "https://github.com/chrisabrams",
+      "contributions": [
+        "code",
+        "doc",
+        "ideas"
+      ]
     }
   ],
   "repoType": "github",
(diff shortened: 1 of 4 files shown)


chrisabrams:
Closing in favor of https://github.com/callstack/linaria/pull/484

PR #489, opened 2019-09-24
docs: add information how to extract className for components

le0pard:
Hello.

We like linaria approach and we would like to use BEM approach with it. We have a simple use case:

```css
.component {
...
}

.component--primary {
...
}
```

We cannot use `className` prop for this purpose because it can contain `className`, which is passed to component on render:

```js
function CoolComponent({ className, style, variant, ...rest }) {
  const classes = classnames(
    className,
    {[`${className}--primary`]: (variant === 'primary')} // will not work, because className === "some-class <component class>"
  )
  return (
    <div className={classes} style={style}>
      {...}
    </div>
  );
}
const StyledCoolComponent = styled(CoolComponent)`
  background-color: tomato;

  &--primary {
    background-color: yellow;
  }
`;

// render

<StyledCoolComponent className="some-class" />
```

Also component can refer to another component, which affects `className` prop:

```js
function CoolComponent({ className, style, muted, ...rest }) {
  const classes = classnames(
    className,
    {[`${className}--muted`]: muted} // will not work
  )
  return (
    <p className={classes} style={style}>
      {...}
    </p>
  );
}
const Paragraph = styled(CoolComponent)`
  font-size: 16px;
  color: #555;

  &--muted {
    color: #ccc;
  }
`;

const Article = styled.article`
  /* when referring to a component, interpolate it as a selector */
  ${Paragraph} {
    font-size: 14px;
    margin: 16px;
  }
`;
```

Our proposed solution is to pass `linariaClassName` prop to wrapped function:  

```js
function CoolComponent({ className, style, linariaClassName, variant, ...rest }) {
  const classes = classnames(
    className,
    {[`${linariaClassName}--primary`]: (variant === 'primary')}
  )
  return (
    <div className={classes} style={style}>
      {...}
    </div>
  );
}
const StyledCoolComponent = styled(CoolComponent)`
  background-color: tomato;

  &--primary {
    background-color: yellow;
  }
`;
```

Thanks.

### DIFF
--- a/docs/BASICS.md
+++ b/docs/BASICS.md
@@ -406,14 +406,17 @@ const StyledCustomCard = styled(CustomCard)`
 `;
 ```
 
-If you want to use linaria classname, which is generated for component, you can use `linariaClassName` prop:
+If you want to use linaria classname, which is generated for component, you can get it as last item from `className` prop:
 
 ```js
-function CoolComponent({ className, style, linariaClassName, variant, ...rest }) {
-  const classes = classnames(
+function CoolComponent({ className, style, variant, ...rest }) {
+  const allClasses = className.split(' ');
+  const linariaClassName = allClasses[allClasses.length - 1];
+
+  const classes = cx(
     className,
-    [`${linariaClassName}--primary`]: (variant === 'primary')
-  )
+    variant === 'primary' && `${linariaClassName}--primary`
+  );
 
   return (
     <div className={classes} style={style}>
--- a/docs/BASICS.md
+++ b/docs/BASICS.md
@@ -406,6 +406,31 @@ const StyledCustomCard = styled(CustomCard)`
 `;
 ```
 
+If you want to use linaria classname, which is generated for component, you can use `linariaClassName` prop:
+
+```js
+function CoolComponent({ className, style, linariaClassName, variant, ...rest }) {
+  const classes = classnames(
+    className,
+    [`${linariaClassName}--primary`]: (variant === 'primary')
+  )
+
+  return (
+    <div className={classes} style={style}>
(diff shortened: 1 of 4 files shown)


le0pard:
Is anyone alive?

thymikee:
cc @Anber

Anber:
I like the approach, but linariaClassName is a non-standard attribute, and it should not be passed to the end component like `div`. I think, we need to find a better solution.

le0pard:
Thanks @Anber . You totally right. 

I decided best way will be to use "order" inside `className` - component `className` always last item in it. So I just add test for this use case, so in future this behaviour will be not broken by some change. And of course add docs about usage.

Hope this will help.

docs/BASICS.md
```diff
@@ -406,6 +406,34 @@ const StyledCustomCard = styled(CustomCard)`
...
 ```
 
+If you want to use linaria classname, which is generated for component, you can get it as last item from `className` prop:
+
+```js
+function CoolComponent({ className, style, variant, ...rest }) {
+  const allClasses = className.split(' ');
+  const linariaClassName = allClasses[allClasses.length - 1];
+
+  const classes = classnames(
```
Anber: We have our own `cx` for composition:
```js
cx(
  className,
  variant === 'primary' && `${linariaClassName}--primary`
);
```

It's just for information. You don't need to fix it here :)

le0pard: Thanks. I will fix this. One moment.

le0pard: @Anber fixed

le0pard:
Big thanks @Anber !

PR #498, opened 2019-10-21
Evaluator as a plugin

Anber:
**Summary**

This is the second part of PR #462. It extracts the old styles evaluator to the plugin and allows to add other evaluators. It also improves the performance of evaluators.

**Test plan**

New tests were added.


src/__tests__/babel.test.ts
```diff
@@ -1,21 +1,18 @@
...
+import { join } from 'path';
+import stripAnsi from 'strip-ansi';
 import serializer from '../__utils__/linaria-snapshot-serializer';
 
 expect.addSnapshotSerializer(serializer);
 
-const transpile = input =>
-  babel.transformAsync(input, {
+const transpile = async (input: string) =>
+  (await babel.transformAsync(input, {
```
thymikee: no need to wrap it in await again, returning a promise is fine

Anber: It is there because of a typecast from `Promise<BabelFileResult | null>` to `Promise<BabelFileResult>`. Any ideas how to do it better?

thymikee: Nah, it's in a test, so I don't mind

Anber:
@thymikee Are we waiting for @satya164’s review or it’s ready for merge?

satya164:
Feel free to merge and publish :D

Issue #420, opened 2019-05-25
Error: Cannot find module 'core-js/modules/es6.regexp.to-string'


egor-a-trubnikov-panov:
I ran this [example](https://codesandbox.io/s/2x93v25p3j) on my computer and I got this error:

`ERROR in ./src/App.js
Module build failed (from ./node_modules/linaria/loader.js):
Error: Cannot find module 'core-js/modules/es6.regexp.to-string'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:613:15)
    at Function.Module._load (internal/modules/cjs/loader.js:539:25)
    at Module.require (internal/modules/cjs/loader.js:667:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/egortrubnikov-panov/Downloads/linaria-demo 2/node_modules/linaria/lib/loader.js:3:1)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
`

how can I fix this?

egor-a-trubnikov-panov:
I added "core-js": "^2.0.0", in package.json and everything worked

skywickenden:
Also having this problem. I'm a first time user to Linaria and was simply following the install instructions.  I added 

    "core-js": "^2.6.8"

to devDependencies. (^2.0.0 has vulnerabilities!)

This looks like https://github.com/callstack/linaria/issues/361 and should be fixed but isn't

c01nd01r:
Looks like `package.json` in master branch and in npm (https://unpkg.com/linaria@1.3.1/package.json) has differences
<details>
  <summary>diff</summary>
  
![img diff](https://user-images.githubusercontent.com/3511312/58386694-1b731300-800c-11e9-9aa6-b5f4e3f1c714.png)
</details>

pbitkowski:
Hi! Thanks for your input. Could you try alpha version?

le0pard:
@pbitkowski thanks. Can you publish 1.4.0-alpha.0 to npm?

Hotell:
[...]

PR #511, opened 2019-12-10
fix: compile babel and server (#420)

oreqizer:
**Summary**

Sets `babel` up for compiling server stuff to target `node@8`. This is to remove the need to install `core-js@2` as a dependency.

**Test plan**

I've checked the build in `lib/` and unnecessary `core-js` requires are gone.

### DIFF
--- a/src/core/.babelrc
+++ b/src/core/.babelrc
@@ -1,0 +1,29 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "browsers": [
+            "last 2 versions",
+            "not ie 11",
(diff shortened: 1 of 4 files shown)


src/core/.babelrc
```diff
@@ -0,0 +1,29 @@
+{
```
satya164: Instead of having duplicate Babel configs you could also set the config based on directory name in the root config https://babeljs.io/docs/en/options#overrides

oreqizer: ok that makes more sense — I will adjust this during the week

jayu:
closing via #569

PR #525, opened 2020-01-11
feat: upgraded corejs to v3

alicanerdogan:
**Summary**
linaria has a core dependency of corejs v2, however corejs v2 is deprecated and it is not supported anymore. To keep up with the other working environment which are up to date, corejs version has been updated to v3 with this plan.

Here is the issue link: https://github.com/callstack/linaria/issues/420

**Test plan**

The PR successfully passes all the tests within repo and it is locally tested to check it's sanity.

### DIFF
--- a/src/__tests__/__snapshots__/preval.test.js.snap
+++ b/src/__tests__/__snapshots__/preval.test.js.snap
@@ -407,7 +407,7 @@ CSS:
    ;
 }
 
-Dependencies: core-js/modules/es6.string.raw
+Dependencies: core-js/modules/es.string.raw
 
 `;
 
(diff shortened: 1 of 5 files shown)


src/__tests__/__snapshots__/preval.test.js.snap
```diff
@@ -407,7 +407,7 @@ CSS:
    ;
 }
 
-Dependencies: core-js/modules/es6.string.raw
+Dependencies: core-js/modules/es.array.map, core-js/modules/es.array.slice, core-js/modules/es.string.raw
```
satya164: We should prolly add `es.array.map` and `es.array.slice` to the `exclude` option in babel config.

alicanerdogan: You are certainly right, I added those two to the `exclude` list

alicanerdogan:
I'm having an issue with a local test, closing PR temporarily.

Issue #517, opened 2019-12-31
[1.4.0-beta.0] A build error


Anber:
This problem was reported by @Guria (#495)

## Description

```
node_modules/react-scripts/node_modules/@babel/runtime/helpers/esm/taggedTemplateLiteral.js:1
export default function _taggedTemplateLiteral(strings, raw) {
^^^^^^

SyntaxError: Unexpected token 'export'
```

Anber:
@Guria could you please provide some details?

Guria:
Will try to dig in and provide reproduction

matiasbuildops:
After upgrading to 1.4.0-beta.0, seeing a the same (or at least, very similar) error message:
```
Module build failed (from ./node_modules/linaria/loader.js):
...\node_modules\react-dnd\dist\esm\index.js:1
export * from './common';

SyntaxError: Unexpected token export
    at new Script (vm.js:83:7)
    at Module.evaluate (...\node_modules\linaria\lib\babel\module.js:200:18)
    at Module.<anonymous> (...\node_modules\linaria\lib\babel\module.js:184:13)
    at Module.require (...\node_modules\linaria\lib\babel\module.js:134:23)
    at C:\Users\matias\code\bops\buildhero-dispatchboard\lib\components\Component2.js:32:17
    at Script.runInContext (vm.js:133:20)
    at Module.evaluate (...\node_modules\linaria\lib\babel\module.js:203:12)
    at Module.<anonymous> (...\node_modules\linaria\lib\babel\module.js:184:13)
    at Module.require (...\node_modules\linaria\lib\babel\module.js:134:23)
    at C:\Users\matias\code\bops\buildhero-dispatchboard\lib\components\Component1.js:3:22
 @ ../lib/components/MainView.js
 @ ../lib/index.js
 @ ./src/App.js
 @ ./src/index.js
```

Not sure if @Guria is also seeing it from linaria *loader*? Seems to me that for some reason, babel is somehow getting confused when getting run through linaria, as both errors seem to be triggered by unrelated modules having ES6 code being included but not getting transpiled by babel?

filenames anonymized a bit; but the use case is basically styled(Component2) (and it works a-ok without that nesting) so seems like a possible regression from #343? Will try to get a better reproduction going as, well.

Anber:
@Guria @matiasbuildops hi guys. Any news here?

Guria:
@anber clean cra + linaria has no such problem. I am still wondering what is needed to add to reproduce this problem

[...]

Issue #512, opened 2019-12-12
Linaria loader break webpack "[hash]" for css and js files


le0pard:
## Description

We are using linaria in our project and found, that if our system build assets in different directories, this break webpack `[hash]` for files. 

```js
module.exports = {
  //...
  output: {
    filename: '[name].[hash].bundle.js'
  }
};
```

Why this happend? Linaria using absolute path for build (extracted from ast):

```json
{"ast":null,"code":"import{styled}from'linaria/react';export var FetchErrorMessageBlock=/*#__PURE__*/styled(\"div\")({name:\"FetchErrorMessageBlock\",\"class\":\"f1bh9ktc\"});export var FetchErrorText=/*#__PURE__*/styled(\"div\")({name:\"FetchErrorText\",\"class\":\"feaenko\"});require(\"/var/www/example/releases/20191212093257/tmp/.linaria-cache/app/javascript/src/app/components/fetchErrorMessage/styles.linaria.css\");","map":{"version":3,"sources":["/var/www/example/releases/20191212093257/app/javascript/src/app/components/fetchErrorMessage/styles.js"],"names":["FetchErrorMessageBlock","FetchErrorText"],"mappings":"AAAA,OAAA,MAAA,KAAA,eAAA,CAEA,MAAO,IAAMA,CAAAA,sBAAsB,CAAA,aAAA,MAAA,CAAA,KAAA,CAAA,CAAA,CAAA,IAAA,CAAA,wBAAA,CAAA,QAAA,UAAA,CAAA,CAA5B,CAIP,MAAO,IAAMC,CAAAA,cAAc,CAAA,aAAA,MAAA,CAAA,KAAA,CAAA,CAAA,CAAA,IAAA,CAAA,gBAAA,CAAA,QAAA,SAAA,CAAA,CAApB,C","sourcesContent":["import {styled} from 'linaria/react'\n\nexport const FetchErrorMessageBlock = styled.div`\n  text-align: center;\n`\n\nexport const FetchErrorText = styled.div`\n  margin-bottom: 5px;\n`\n"]},"metadata":{},"sourceType":"module"}
```

As we can see, here two issues, which break source map hash (which influence on files) and js files hash. First part this:

```js
require(\"/var/www/example/releases/20191212093257/tmp/.linaria-cache/app/javascript/src/app/components/fetchErrorMessage/styles.linaria.css\");
```

[...]

PR #529, opened 2020-01-21
Fixes relative imports failing in workspaces when cwd() is not workspace root

matiasbuildops:
First off; consider this a someting between a PR and a bug report, as I'm not really an expert on babel, module resolution, yarn workspaces, linaria internals or practically any of the things this PR touches upon. It's more along the lines of a minimal fix that *seems* to work based on my limited understanding and testing. 

I really have no idea how all of these bits actually work together (for example, when and on what files does the transform->transformFromAstAsync actually get called? doesn't seem to be all of them..) so keep that in mind!

**Background**

Initially, a PR (#522) was created to support creating cache directories to be aware of Yarn workspaces. This introduced a new root directory constant in `loader()`, in order to keep cache directories in their relative packages and ultimately to allow for components in separate packages to work without a hitch.

There was an older open issue (#512) with hashes not staying constant across builds that was fixed by passing relative paths (to the root, eg. the workspace root) to webpack, as recommended by webpack. This also necessitated changing the filename passed on to transform() ([to be relative to root](https://github.com/callstack/linaria/blob/29ffee9a02b7a9286f961acc8f3e0e0e71592f16/src/loader.ts#L65)), I'm guessing so that the filenames match what webpack gets? Not quite sure of this, to be honest!

However, the problem with this arises with relative imports when cwd() does not match the workspaces root (eg. when running a script from a package directory / package.json). The linaria transform function [calls](https://github.com/callstack/linaria/blob/a3ec9a96bbe1f1de50365f04f5ce971d7fcf8114/src/transform.js#L71) babel `transformFromAstAsync`, passing in options the filename (which is relative to root eg workspace root eg. includes the package name). Crucially, this call to babel does not take into account the pluginOptions passed in, so will not respect user-set presets or other babel configuration.

(The configuration passed in here, I think, also is the root cause of #517 as basically it iss expecting any imported file to already be babelized which it might not be, if it's imported from node_modules along with, as is pretty common, having an ignore for node_modules for the first babel pass? I'm not 100% on the babel/linaria/webpack semantics involved so might be wildly off base here)

The first thing the babel `transformFromAstAsync` does is to parse the options [calling](https://github.com/babel/babel/blob/9bc04baeb5bca0025a54fdd757df5e230c3ce060/packages/babel-core/src/transform-ast.js#L32) `loadConfig` which points to [`loadFullConfig`](https://github.com/babel/babel/blob/fa975bf7cd2b9054faaff107a79e41dcaad305b1/packages/babel-core/src/config/full.js#L59) that first calls [`loadPrivatePartialConfig`](https://github.com/babel/babel/blob/9bc04baeb5bca0025a54fdd757df5e230c3ce060/packages/babel-core/src/config/partial.js#L66), that [takes the filename and appends it to cwd](https://github.com/babel/babel/blob/9bc04baeb5bca0025a54fdd757df5e230c3ce060/packages/babel-core/src/config/partial.js#L94) when it parses the config, causing the final resolved filename to be `[package]/[package]/[file]` instead of the `[package]/[file]` that would be expected, that causes a rather grim-looking build error with very little indicating the actual problem: 

(the little is the crucial part, although it's a bit easy to miss within the paths: Error: Can't resolve './file' in 'workspace-root-cwd\\**package\package**\src')

```
ERROR in ./src/main.js
Module build failed (from ../node_modules/linaria/loader.js):
workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:147
                if (err) throw err;
                         ^

Error: Can't resolve './file' in 'workspace-root-cwd\package\package\src'
    at doResolve (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:209:21)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:15:1)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:27:1)
    at resolver.doResolve (workspace-root-cwd\node_modules\enhanced-resolve\lib\DescriptionFilePlugin.js:67:43)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn1 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:16:1)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:27:1)
    at resolver.doResolve (workspace-root-cwd\node_modules\enhanced-resolve\lib\DescriptionFilePlugin.js:67:43)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn1 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:16:1)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:15:1)
    at fs.stat (workspace-root-cwd\node_modules\enhanced-resolve\lib\DirectoryExistsPlugin.js:27:15)
    at SyncAsyncFileSystemDecorator.stat (workspace-root-cwd\node_modules\enhanced-resolve\lib\SyncAsyncFileSystemDecorator.js:15:12)
    at resolver.getHook.tapAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\DirectoryExistsPlugin.js:22:9)
    at AsyncSeriesBailHook.eval [as callAsync] (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:7:1)
    at Resolver.doResolve (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:282:16)
    at resolver.getHook.tapAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\TryNextPlugin.js:19:14)
    at _next0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:8:1)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:30:1)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn5 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:16:1)
    at hook.callAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\Resolver.js:285:5)
    at _fn1 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:16:1)
    at fs.stat (workspace-root-cwd\node_modules\enhanced-resolve\lib\FileExistsPlugin.js:24:14)
    at SyncAsyncFileSystemDecorator.stat (workspace-root-cwd\node_modules\enhanced-resolve\lib\SyncAsyncFileSystemDecorator.js:15:12)
    at resolver.getHook.tapAsync (workspace-root-cwd\node_modules\enhanced-resolve\lib\FileExistsPlugin.js:20:8)
    at _next0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:8:1)
    at _fn0 (eval at create (workspace-root-cwd\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:30:1)
    at forEachBail.withIndex (workspace-root-cwd\node_modules\enhanced-resolve\lib\SymlinkPlugin.js:42:36)
    at args (workspace-root-cwd\node_modules\enhanced-resolve\lib\forEachBail.js:64:14)
    at fs.readlink (workspace-root-cwd\node_modules\enhanced-resolve\lib\SymlinkPlugin.js:38:8)
    at SyncAsyncFileSystemDecorator.readlink (workspace-root-cwd\node_modules\enhanced-resolve\lib\SyncAsyncFileSystemDecorator.js:48:12)
```

A repo with a reproducible demo can be found at https://github.com/matiasbuildops/linaria-workspace-root-cwd

**Summary**

This PR changes the filename passed in options to transform() to be relative to cwd() rather than the workspace root. This is because if we're in a workspace environment and cwd() is not the same as workspace root, transform() passing this filename (with the package name) on to babel causes babel the package name to be repeated in the path and further resolution will then fail.

**Test plan**

Testing would probably involve mocking `find-yarn-workspace-root` to return the parent directory or something along those lines. 

However, my current development environment is set up on Windows and just getting a commit turned out to be quite a pain with git.autocrlf = true causing linting to fail on every line and many of the snapshot tests fail even on a fresh checkout due to path separators being different and it showing up in many of the snapshots. In addition, many of the hashes in the snapshots have changed after the change, but I guess this is working as designed? No other test failures except these two types are reported. Happy to try to write that test when I find some time to get more comfortable with the testing side of linaria.

Updating snapshots does make all tests pass.

Also tested manually that rebuilding the same source does not change the hashes of the files webpack builds, as seemed to be the problem in #512.

**Questions**

As I'm really not sure of all the implications of this change, this might not be the best way to solve it. Other options I can think of would include at least:

- creating the relative to cwd() path to pass on to babel only in transform() rather than loader() (should reduce other implications?)
- allowing root to be configurable by the user (from webpack config) (would allow for more exotic configurations, too)
- allowing the user to pass on options to the babel that is called from transform() (I think this could allow #517 to be fixed, too?) (perhaps we could just pass in pluginOptions.babelOptions?)


Anber:
Hi @matiasbuildops
Thank you for PR. May I ask you to rebase your commit on top of `callstack:develop`? :)

Anber:
Sorry @matiasbuildops, it was the wrong decision to rebase develop before merging your PR. I've manually cherry-picked your commit.

matiasbuildops:
Thanks @Anber , sorry for not getting back to this earlier!

PR #533, opened 2020-01-23
chore(ci): run unit tests on Node v10 and v12

### LINKED ISSUE #206: CI: also test Node.js 8 and 10
We should also test Node.js 8 and 10.

koba04:
**Summary**

This fixes #206 
Currently, The CI runs on Node v10. 
The current Node LTS versions are v10 and v12, so we should test on the versions.
https://nodejs.org/en/about/releases/

This PR changes the Node version of lint tasks to v12 (current LTS version) and run unit tests on v10 and v12.

**Test plan**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,75 +1,110 @@
-version: 2
-defaults: &defaults
-  docker:
-    - image: circleci/node:10
-  working_directory: ~/linaria
-jobs:
-  install-dependencies:
-    <<: *defaults
-    steps:
-      - checkout
-      - attach_workspace:
-          at: ~/linaria
-      - restore_cache:
-          keys:
-            - dependencies-{{ checksum "package.json" }}
-            - dependencies-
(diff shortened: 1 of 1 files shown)


koba04:
I've changed CI jobs so the checks(`install-dependencies` and `unit-tests`) have never finished...

Anber:
Hi @koba04 
Thank you for your PR!

Did I understand correctly that `install-dependencies` will be called independently for each job instead of one call in the current version?

koba04:
@Anber 

> Did I understand correctly that install-dependencies will be called independently for each job instead of one call in the current version?

Yes, the `install-dependencies` task runs each job, but it doesn't make the CI job slower because they run as parallel.
But I understand your concern, so I've separated `install-dependencies` as a job.

https://circleci.com/workflow-run/d1f59730-3931-4c8e-aa6a-880fa17c34b8

koba04:
Thanks!

PR #556, opened 2020-03-20
Bump @emotion/is-prop-valid

Pytal:
The latest `@emotion/is-prop-valid` package contains the most up to date valid props so that all valid props such as `loadng` for `img` tags are passed to the DOM as expected.

**Summary**

Linaria would filter out previously invalid props but now valid props such as `loading` for `img` resulting in unexpected output where the loading prop would not be passed to the DOM.

**Test plan**

1. Add `loading` prop to an `img` tag
2. Bundle code and load webpage
3. Inspect the `img` tag and see that the `loading` prop exists on the DOM


Anber:
Hi @Pytal 
Thank you for your PR!

Can you please commit `yarn.lock` as well?

jayu:
Stale PR, I'm closing

PR #557, opened 2020-03-21
fix lock files

jancassio:
**Summary**

Just updated the `yarn.lock` by issues to use the right version of `core-js`.
Even after install the latest version of `core-js` in main project depenencies, Linaria still using `core-js@2x`.

**Test plan**

No test required, for this change, at all, I don't know if we can test it.


jayu:
core-js issue resolved here #569

Issue #554, opened 2020-03-07
Support for the "dash" symbol in the CSS identifiers


klimcode:
## Dash is a valid symbol frequently used in CSS classnames.

Would be great to include this symbol in the sanitization regex here: "linaria/src/babel/utils/toValidCSSIdentifier.ts", so the regex will be like this:
`/[^-_0-9a-z]/gi`

## Motivation

Currently it's impossible to use this Slug "[title]-[hash]". The dash is converted to underscore.

The difference is significant: in the DevTools, double click on a classname "class-abcde" will select only the first part, so it's easy to copy the original classname.
But doubleclick on the "class_abcde" will select the scoped name which forces me to manually remove the hash from selection or select the classname not by doubleclick, but dragging the cursor over the original classname which is much slower and irritating.

soluml:
To add to this issue, there are actually thousands of valid characters beyond the regex above. Mathias Bynens has a [great write up about escaping normally invalid characters for CSS class names](https://mathiasbynens.be/notes/css-escapes) and in that same article he discusses many of character ranges that can be used, including emojis which don't require any escaping at all.

Unless there's a programmatic reason to have Linaria sanitize class names, I think it'd be better to just not sanitize them at all. I think it's fair to trust the developer to use the slug name correctly.  If you do need to sanitize certain characters, then I think it's important to document which characters will be filtered out or transformed and to what.

Finally, at the very least, I too would love to get dashes back 👍

jayu:
Thanks for the research.  I would highly encourage you, folks, to check if changing this regexp does not break the test (it probably will break some snapshots) and what's more important -  builds.  I don't know why it is so strict, maybe there was a reason. From a maintainer perspective, the issue does not have very high priority but is reasonable so I look forward to community contributions :D  

Please check [how you can link](https://github.com/callstack/linaria/blob/master/CONTRIBUTING.md) the development version of Linaria into your existing project

jayu:
Lol, didn't notice the PR is already there 😅 Thanks!

soluml:
PR #564  has merged, so I think this can be closed once available.  Thanks everyone!

PR #564, opened 2020-04-01
Expand the valid CSS Identifier method to allow more valid characters in class names

soluml:
**Summary**

The hyphen (or dash) character is a valid character for CSS class names (under most circumstances) and does not need to be escaped. Characters from the Latin-1 Supplement unicode character range through the Specials unicode range are also valid and do not need to be escaped.

This PR expands the regex used to filter out invalid characters to allow these.

**Test plan**

There are no tests for this method currently. However, we can observe that these characters are valid for CSS class names by [looking at the spec](https://www.w3.org/TR/CSS2/syndata.html#characters). You can [read more information about the additional unicode ranges added](https://jrgraphix.net/research/unicode_blocks.php), and see a read a fantastic write up about [constructing class names here](https://mathiasbynens.be/notes/css-escapes).

This PR will address issue #554.


src/__tests__/__snapshots__/babel.test.ts.snap
```diff
@@ -1,559 +1,559 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Anber, line 1: It looks like your snapshot has a different line ending symbol. Could you please check it?

jayu:
Let me check if this does not break my example project builds ^^ I think we should add example project building to our pipelines

Anber:
@jayu we can use http://linaria.now.sh/ as an example project :)

jayu:
@Anber It's a great idea but we might want to have better separation of concerns. I opened an issue for that #565

jayu:
Just checked and it works for me. @soluml  please address @Anber comment and we are ready to go.

src/__tests__/__snapshots__/babel.test.ts.snap
soluml: Yep, my bad! Editor didn't recognize this as a LF project. Updated!

soluml:
@jayu Line Endings for the .snap file have been resolved.

PR #567, opened 2020-04-02
docs: new issue templates

jayu:
**Summary**

I would like to extend and improve issue templates to fit better Linaria use cases and work with new labels

**Test plan**

I will merge it to my fork and test if labels work and if lines of comments are not broke in wrong places
**Edit**: everything works fine.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Report a bug 🐛
+about: Report a reproducible bug or reproducible regression.
+labels: 'bug 🐛'
```
thymikee, line 4: It's better to go with a label `bug report`, because they're often not bugs. Also it's worth to add something like `needs repro` and `needs triage` by default. Check how we do it at Jest: https://github.com/facebook/jest/tree/master/.github/ISSUE_TEMPLATE

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+
+<!--
+Especially include
+- version of Linaria you are using
+  - check exact version in lock file
+- used bundler eg. webpack with it's version
+- platform/framework eg. react or next.js
+- node version
+- operating system with version
+-->
```
thymikee, line 21: Make it a list, not a comment:
```
- Linaria version:
- Bundler (+ version): 
- Node.js version:
- OS: 
```
You can also leverage https://github.com/tabrindle/envinfo

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,9 @@
+---
+name: Question ❓
+about: Choose if you need help.
+labels: 'question ❔'
```
thymikee, line 4: You should focus on making this an issue tracker, not a help forum. Redirect people to Stack Overflow or a Discord channel. See how RN is doing it:

image

.github/ISSUE_TEMPLATE/bug_report.md
jayu: Good idea!

.github/ISSUE_TEMPLATE/question.md
jayu: Hmm, people here are often asking "Does linaria works with.."  Or "How I can me it work with" 
I think in current status of the project those are actually useful questions because we can collect more feedback and identify needs. But I will rephrase `about` to be more informative

.github/ISSUE_TEMPLATE/bug_report.md
jayu: Make sense. We understand `bug reports` without `needs tirage` as real bugs

jayu: Or maybe two labels are actually more useful

PR #569, opened 2020-04-03
chore: change babel config and remove need for core-js dependency

### LINKED ISSUE #561: Solve issue with core-js, check if we can remove or at least upgrade.
## Motivation

There are a few issues related to core-js. They are built errors and build performance issues.
We should check how we can address them. We need to investigate
- [ ] if we can remove `core-js` dependency
- [x] if we can upgrade `core-js` to most recent version
- [ ] check what stop of the maintenance of `core-js` means for us in terms of upgrading/removing a dependency https://github.com/zloirock/core-js/issues/548#issuecomment-494112872 and https://github.com/zloirock/core-js/issues/767

## Related Issues

#420 #468

jayu:
## Motivation

It closes #561

## Summary

I refactored `react/styled.ts` to not require core-js. 

I Changed default `babel-preset-env` target to node 10 ( kudos goes to @oreqizer for #420) to remove core-js dep in non-browser packages. 

I struggled with tests, it turns out that we have to provide browser-like `babel-preset-env` to not break them. There are 3 tests that fail without this setup, I give up on trying to figure out why. 

## Test plan

green CI

I'm still not sure about those config changes.  It should be ok for current users because we excluded all introduced core-js polyfills. But if it eventually breaks something, we still going to release it as `2.0` so we are more or less safe 🤷‍♂


babel.config.js
```diff
@@ -0,0 +1,69 @@
...
+       * only src/react and src/core are targeted to be run in the browser
+       */
+      test: /src\/((react)|(core))\//,
+      presets: [
+        '@babel/preset-react',
+        [
+          '@babel/preset-env',
+          {
+            targets: {
+              browsers: [
```
thymikee, line 26: for `test` env we should target `node` instead of browsers

babel.config.js
```diff
@@ -0,0 +1,69 @@
...
+       * we use the same config for src/__tests__ and src/__fixtures__ to not break existing tests
+       */
+      test: /src\/((__tests__)|(__fixtures__))\//,
+      presets: [
+        '@babel/preset-react',
+        [
+          '@babel/preset-env',
+          {
+            targets: {
+              browsers: [
```
thymikee, line 55: for `test` env we should target `node` instead of browsers

babel.config.js
```diff
@@ -0,0 +1,69 @@
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          node: '8',
```
thymikee, line 7: Node 8 is already EOL, how about going Node 10?

jayu: Hmm, that's might be a point.

babel.config.js
jayu: Unfortunately, some tests are failing when we switch the target to `node`. I mentioned that in the PR description.

thymikee:
Please note that technically, changing Node.js target to 10 is a breaking change

src/__tests__/detect-core-js.test.js
```diff
@@ -0,0 +1,26 @@
...
+      resolve(output);
+    });
+  });
+};
+
+it('Ensures that package do not include core-js dependency after build', async () => {
+  const packageJSON = require('../../package.json');
+  const buildScript = packageJSON.scripts.build;
+
+  const proc = cp.exec(`NODE_ENV=debug ${buildScript}`, {
```
jayu, line 19: I wonder if we should care about windows users and set env variable like that.

thymikee: You can pass env through options like `{env: {NODE_ENV: 'debug'}}`
https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

jarodpeachey:
@jayu When will this change be released? I'm currently getting errors related to core-js and would love to use the new version

jayu:
@jarodpeachey I will try to release it today, we need to decide internally whether it should be still 1.x-beta or 2.0-alpha

PR #570, opened 2020-04-04
fix(evaluator): forces to use cjs for plugin-transform-runtime

Anber:
## Motivation

Evaluators tried to use default Babel config which is could be configured for using ES-modules that can cause injecting the wrong versions of babel-helpers during evaluation stages. (#517)

## Summary

This PR explicitly adds `@babel/plugin-transform-runtime` to pre-evaluation transforms.


jayu:
I didn't manage to make code review :(

Issue #571, opened 2020-04-09
Add a third argument to the preprocessor option that gives us the file name


soluml:
## Describe the enhancement
In the webpack linaria/loader, we have a [preprocessor option](https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#options) which gives us the ability to further augment the text before passing it down the line. It would be very helpful to have a 3rd argument which features the output file name for this file.

## Motivation

This would allow us to conditionally make changes, like, adding a Sass `@use` at the top of the file to include mixins and or variables.

Currently, we have to work around it doing something like so:

```
{
  loader: 'linaria/loader',
  options: {
    extension: '.linaria.scss',
    preprocessor: (selector, cssText) => `
      @import 'Styles/helpers';
      ${selector} { ${cssText} }
    `,
    sourceMap: stage.startsWith('develop'),
  },
}
```

This works, but results in importing the file over and over again a lot. With a third argument, we could determine which file we're in and conditionally add `@use`/`@import`'s as needed.

jayu:
@soluml As I understand, Inside preprocessor callback you want to access the filename from where styles were extracted? Or rather the path. So the use case might look like
```
 preprocessor: (selector, cssText, filePath) => filePath === 'foo' ? `
      @import 'Styles/helpers';
      ${selector} { ${cssText} }
    ` : 'sorry, co css',
```

soluml:
Yea, exactly. Sorry I wasn't clear. Below is something I'd do if this was implemented:

```
// Somewhere in the file
const filePathMap = new Map();

// Loader Options
preprocessor: (selector, cssText, filePath) => {
  const imports = filePathMap.get(filePath) ? "" : "@use 'Styles/helpers';";  
  filePathMap.set(filePath, true);
  return `${imports} ${selector} { ${cssText} }`;
}
```
Basically, if this file has been written to already, don't add [Sass's @use](https://sass-lang.com/documentation/at-rules/use) as it must be at the top of each file.

soluml:
I opened up a PR for this issue: #572

satya164:
@soluml not sure if keeping external map is something you should do. files can be rebuilt at a future point. your external cache won't be consistent with the actual output.

soluml:
[...]

PR #572, opened 2020-04-09
Pass the Webpack options.filename to the preprocessor function.

soluml:
## Motivation

Solves issue #571.

## Test plan

Updated types to reflect the addition of a third argument. No real functionality changes, just passing in additional data which was provided by the bundler.

### DIFF
--- a/src/transform.ts
+++ b/src/transform.ts
@@ -133,7 +133,11 @@ export default function transform(code: string, options: Options): Result {
     });
 
     // Run each rule through stylis to support nesting
-    cssText += `${preprocessor(selector, rules[selector].cssText)}\n`;
+    cssText += `${preprocessor(
+      selector,
+      rules[selector].cssText,
+      options.filename
+    )}\n`;
   });
 
   return {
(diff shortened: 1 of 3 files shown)


src/transform.ts
```diff
@@ -133,7 +133,11 @@ export default function transform(code: string, options: Options): Result {
     });
 
     // Run each rule through stylis to support nesting
-    cssText += `${preprocessor(selector, rules[selector].cssText)}\n`;
+    cssText += `${preprocessor(
+      selector,
+      rules[selector].cssText,
+      options.filename
```
satya164, line 139: Can you make this an object instead? It will be more futureproof if there's need for more metadata

soluml: Sure! That's a good idea. Instead of just filename I'll pass the entire options object as a 3rd argument.

soluml: Done

jayu:
I treat this PR as stale, I will reopen if @soluml decide to update!

PR #574, opened 2020-04-14
chore: setup super-laberer-action for automated issue labeling

jayu:
## Motivation

Recently I added a bunch of new labels to organize issues management.
I wanted to automate it a bit, so I integrated `super-laberer-action` https://github.com/IvanFon/super-labeler-action/

## Summary

However, I'm using my fork, because mine and @jbina PR's are not yet merged. We can migrate later. 

It basically adds labels based on matched keywords, mostly case-insensitive. 

It contains labels list extracted from original list created in linaria using this gist https://gist.github.com/jayu/b80c5f28a79a810a177d5ee97d7ab258

It should work out-of-the-box, maybe we will need to enable GitHub actions if not enabled yet.

**If you have an idea for more/better keywords please leave a comment.**

I didn't cover all labels, eg. didn't cover `platform: react` because half of the issues have `react` in their content. 

My fork of issue laberer has undocumented feature of 
- adding fallback label (`needs: tirage`) if bot didn't manage to add enough labels
- disabling autolabeling by adding a specific label (`skip autolabeling`) eg. if someone added keyword `webpack` but issue is not related to webpack, we can disable autolabeling to not have to fight with the bot :P 

## Test plan

You can test how it works on my fork https://github.com/jayu/linaria/issues


jayu:
It's funny that it fails, it may be because of missing config file at `master`. I could check to debug logs if someone with admin access would set `ACTIONS_STEP_DEBUG ` env to `true` :)

.github/labels.json
```diff
@@ -0,0 +1,429 @@
...
+  },
+  "issue_fallback": {
+    "fallbackActivationValue": 2,
+    "labels": [
+      "needs-triage"
+    ]
+  },
+  "pr": {},
+  "skip_labeling": "skip-labeling"
+}
```
thymikee, line 429: extra newline?

jayu:
@thymikee Labels defined in `labels.json` are the same labels that exist in the repo setup. So I'm not adding any new labels, maybe with exception fir `skip-labeling`

.github/labels.json
jayu: WDYM?

thymikee: it's common to include an extra newline at the end of the files, so that git doesn't report "no newline..." 
image

thymikee:
> Labels defined in labels.json are the same labels that exist in the repo setup. So I'm not adding any new labels

Yea, but normally they're not stored in the repo and you can still use them to label issues. What I was curious about is if this action needs such `labels.json`, but now I read the doc of this action and it's clear it's needed. Sorry for confusion

PR #576, opened 2020-04-16
docs: start 2.0 migration guide

jayu:
## Motivation

I wanted to inform users on current status of the project since for almost a year there were `1.4.0-beta` releases without a stable version and now we are starting `2.0.0` which might be confusing. 

## Summary

Updated Readme, added Migration guide.

## Test plan

n/a


jayu:
Maybe instead of a change in the readme, it would be more reasonable to create a pinned issue?

Issue #575, opened 2020-04-14
SyntaxError: Unexpected token 'export'


SidneyLann:
- Linaria version: 1.1 to 1.4
- Bundler (+ version):
- Node.js version: 13
- OS: win7

## Description
Module build failed (from ./node_modules/_linaria@1.4.0-beta.6@linaria/loader.js):
D:\DEV\client-taro\node_modules\_@tarojs_components@3.0.0-beta.3@@tarojs\components\mini\index.js:1
(function (exports) { export const View = 'view'
                      ^^^^^^

SyntaxError: Unexpected token 'export'

## Reproducible Demo
use taro 3 to build a mini program for wechat should reproduce the issue

SidneyLann:
export const View = 'view'
export const Icon = 'icon'
export const Progress = 'progress'
export const RichText = 'rich-text'
export const Text = 'text'
export const Button = 'button'
export const Checkbox = 'checkbox'
export const CheckboxGroup = 'checkbox-group'
export const Editor = 'editor'
export const Form = 'form'
export const Input = 'input'
export const Label = 'label'
export const Picker = 'picker'
export const PickerView = 'picker-view'
export const PickerViewColumn = 'picker-view-column'
export const Radio = 'radio'
export const RadioGroup = 'radio-group'
export const Slider = 'slider'
export const Switch = 'switch'
export const CoverImage = 'cover-image'
export const Textarea = 'textarea'
export const CoverView = 'cover-view'
export const MoveableArea = 'moveable-area'
export const MoveableView = 'moveable-view'

SidneyLann:
Above is the content of the file node_modules\_@tarojs_components@3.0.0-beta.3@@tarojs\components\mini\index.js

SidneyLann:
https://taro-docs.jd.com/taro/next/docs/next/css-in-js.html
I configure the env by the url

satya164:
You need to remove this module from ignore list. See `ignore` https://github.com/callstack/linaria/blob/master/docs/CONFIGURATION.md

two-legs:
Having the same issue. 
I have CRA project with ejected config and es6-package in `node_modules`. 
1. I added path to package (atomsLib) into include section in the webpack config: 
```javascript
            {
              test: /\.(js|mjs|jsx|ts|tsx)$/,
              include: [paths.appSrc, paths.atomsLib],
              use: [
                {
                  loader: require.resolve('babel-loader'),
                  options: {
                    customize: require.resolve(
[...]

PR #578, opened 2020-04-17
fix(module): fixes a conflict between old ignore option and the new one

Anber:
This PR fixes a conflict between `ignore` options (#575).


Anber:
@jayu do we need a second PR, or we can just cherry-pick changes?

PR #580, opened 2020-04-19
Fix empty CSS variables

ImedAdel:
## Motivation

Not passing props to a component results in a list of empty CSS variables. For instance, the following component:

```tsx
export const Box = styled.div<BoxProps>`
	display: ${props => props.display || "flex"};
	background: ${props => props.background || ""};
	background-color: ${props => props.backgroundColor || ""};
	flex-wrap: ${props => props.flexWrap || ""};
	flex-direction: ${props => props.flexDirection || ""};
	margin: ${props => props.margin || ""};
	padding: ${props => props.padding || ""};
	height: ${props => props.height || ""};
	width: ${props => props.width || ""};
	border: ${props => props.border || ""};
	border-radius: ${props => props.borderRadius || ""};
	border-width: ${props => props.borderWidth || ""};
	border-color: ${props => props.borderColor || ""};
`;
```

When later used without specifying all of the props, results in a list of empty CSS variables such as the following:

2020-04-19 20_55_35-Centige

## Summary

This fix would skip any `null`, `undefined`, or empty string variable.

### DIFF
--- a/src/react/styled.ts
+++ b/src/react/styled.ts
@@ -98,7 +98,7 @@ function styled(tag: any): any {
 
           warnIfInvalid(value, options.name);
           // skip empty values
-          if(value !== null && typeof value !== undefined && value !== "") {
+          if (value !== null && typeof value !== undefined && value !== '') {
             style[`--${name}`] = `${value}${unit}`;
           }
         }
--- a/src/react/styled.ts
+++ b/src/react/styled.ts
@@ -97,8 +97,10 @@ function styled(tag: any): any {
           const value = typeof result === 'function' ? result(props) : result;
 
           warnIfInvalid(value, options.name);
-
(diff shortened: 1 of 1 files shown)


satya164:
The reason it's not doing that is because it's most likely a bug/typo when a prop is undefined/null. There's even a warning for that. In addition `border-radius: ;` is not a valid CSS.

If your props can be `undefined`, you should specify some default values to use instead using `defaultProps`.

ImedAdel:
@satya164 Yup. That's what I was initially doing. However, Emotion, for example, skips undefined values. I think that it can be cleaner for some use cases.

jayu:
I agree with @satya164 and I think it will be easier to debug issues if we left the variables. But thanks @ImedAdel for your contribution anyway!

PR #585, opened 2020-04-22
docs: add headers to files, more debug statements and instructions how to debug

jayu:
## Motivation

Since codebase looks mysterious for the newcomers, I added headers containing high-level descriptions of file purpose and added more debug statements to easier analyze how Linaria works.

## Summary

- added debug statements
- added comment headers

## Test plan

Read comments and check for logic and grammar mistakes.


src/babel/utils/logger.ts
```diff
@@ -60,3 +60,9 @@ export const debug = log.bind(null, 'debug');
 export const info = log.bind(null, 'info');
 export const warn = log.bind(null, 'warn');
 export const error = log.bind(null, 'error');
+
+export const debugNewLine = () => {
+  if (process.env.DEBUG === 'linaria*') {
```
Anber, line 65: It won't work for cases like `DEBUG=linaria:lazy-deps:*`

src/babel/extract.ts
```diff
@@ -93,14 +104,15 @@ export default function extract(_babel: any, options: StrictOptions) {
           state.index = -1;
           state.dependencies = [];
           state.replacements = [];
+          debugNewLine();
+          debug('start extraction for:', state.file.opts.filename);
```
Anber, line 108: The first arg of `debug` should be a namespace which can be used for logs filtering.

src/babel/utils/logger.ts
jayu: Good point, I will fix it

jayu: I decided to remove it at all, I initially used it to split logs for particular files, but it seems to not be reasonable to debug more than one file at once.

src/babel/extract.ts
jayu: fixed

PR #586, opened 2020-04-22
feat: more verbose report of evaluation error

### LINKED ISSUE #559: Add error logging to webpack plugin
## Describe the feature

It would be very useful to have better error logging from webpack plugin

## Motivation

Currently, when a build fails due to evaluation error, we get an actual error from the evaluation. But instead of a real file name where an error occurred, we get the name of the file that was entry point for evaluation. See #399 for example.

We should print a caught error here, between `try` and `finally` blocks https://github.com/callstack/linaria/blob/672b0bfcedfba1c2a5a5e6f5bfae2d63bfeee241/src/loader.ts#L66-L83

### LINKED ISSUE #581: Improve error logging once Linaria is used in non-static context
## Describe the enhancement

We should print more informative error message when Linaria evaluation fails.

## Motivation

Users, especially new users are confused when they cannot use some kind of syntax, and error message just says 

>An error occurred when evaluating the expression [...] make sure you do not use any Node or Browser specific API. 

## Possible implementations

We should list possible reasons for the error like
- css/styled used inside a loop
- css/styled used inside a function
- css/styled used inside react component
- more 

I'm not sure if we can reason about the context of the error, meaning if we can check if it was used inside function/loop to produce a more tailored error. 

## Related Issues

The goal of this issue is better to error logging so it might be related to #559

jayu:
## Motivation

It closes #581 and closes #559 

## Summary

The evaluation may fail in two ways
- it is not possible to resolve the value because it's dependencies are not available in a static context
- some part of code before actual dependencies was invalid and crashes the vm

I added more information to the first error
Screen Shot 2020-04-22 at 14 28 25

I added a stack trace to the second type of error. However, it is not ideal, because line numbers will not match in most cases. I don't know how to improve it, adding some source maps might be complicated. I assume it might not be worth fixing that because those errors will be very rare with the new shaker evaluator. Currently, the error in web back is more verbose thanks to that:
Screen Shot 2020-04-22 at 17 32 36

before it skipped the original stack trace
Screen Shot 2020-04-22 at 17 32 02

## Test plan

Green CI

### DIFF
--- a/src/__tests__/__snapshots__/preval.test.ts.snap
+++ b/src/__tests__/__snapshots__/preval.test.ts.snap
@@ -765,7 +765,20 @@ Dependencies: NA
 `;
 
 exports[`extractor non-hoistable identifiers 1`] = `
-"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: days is not defined. Make sure you are not using a browser or Node specific API.
+"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: 
+
+  > days is not defined. 
+
+  Make sure you are not using a browser or Node specific API and all the variables are available in static context.
+  Linaria have to extract pieces of your code to resolve the interpolated values.
+  Defining styled component or class will not work inside:
+    - function,
+    - class,
+    - method,
+    - loop,
+  because it cannot be statically determined in which context you use them.
+  That's why some variables may be not defined during evaluation.
+      
    7 | export const Title = styled.h1\`
    8 |   &:before {
 >  9 |     content: \\"\${days}\\"
@@ -857,7 +870,20 @@ Dependencies: @babel/runtime/helpers/interopRequireDefault, react
 `;
 
 exports[`extractor throws codeframe error when evaluation fails 1`] = `
-"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: This will fail. Make sure you are not using a browser or Node specific API.
+"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: 
+
+  > This will fail. 
+
+  Make sure you are not using a browser or Node specific API and all the variables are available in static context.
+  Linaria have to extract pieces of your code to resolve the interpolated values.
+  Defining styled component or class will not work inside:
+    - function,
+    - class,
+    - method,
+    - loop,
+  because it cannot be statically determined in which context you use them.
+  That's why some variables may be not defined during evaluation.
+      
   4 | 
   5 | export const Title = styled.h1\`
 > 6 |   font-size: \${foo()}px;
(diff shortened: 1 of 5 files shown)


src/__tests__/__snapshots__/preval.test.ts.snap
```diff
@@ -765,7 +765,20 @@ Dependencies: NA
...
 
 exports[`extractor non-hoistable identifiers 1`] = `
-"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: days is not defined. Make sure you are not using a browser or Node specific API.
+"<<DIRNAME>>/source.js: An error occurred when evaluating the expression: 
+
+  > days is not defined. 
+
+  Make sure you are not using a browser or Node specific API and all the variables are available in static context.
+  Linaria have to extract pieces of your code to resolve the interpolated values.
+  Defining styled component or class will not work inside:
```
Anber, line 774: Instead of showing on big error here, we can easily modify `extract.ts` in order to check that `TaggedTemplateExpression` is placed on the top level of a file.

jayu: It makes some sense, but the evaluation will work if the user will not use any dependencies. So we should do it only for lazy dependencies, do I understand correctly?

jayu: And it will work in some cases like this
```js
const a = () => {
  const color = 'red'
  return styled.h1`
    color: ${color}
  `
}
```

Anber: It will work but in an unpredictable way. For example, what if a referenced variable is defined in both scopes?

jayu: Good point. Do you mean to remove the possibility to use Linaria outside the top-level scope of the file? 
That would be a quite huge breaking change but in terms of stability, it would improve much. I think we should discuss it more. 
What is your idea to handle that, if we check that `TaggedTemplateExpression` is out of top-level scope, what then? until it does not raise an error we shouldn't do anything to not introduce breaking change.

Anber: It's the same situation as with hooks in React: is it possible to use it in loops? Yes, it is. Is it safe? No.
There is only one safe case when you can use Linaria out of top-level scope: all interpolated variables are defined in a top-level scope. But that means that you can easily move styled/css call to the top-level. So I would prefer to throw an error if out of top-level case is detected.
For usage inside components, we can probably introduce some `useStyles` hook and, if I properly remember, it was even discussed here.

jayu: Throwing an error is a good idea, but Let's first discuss it internally. We can merge this error reporting to both `2.x` and `1.4.x` and then If we decide to restrict usage to top-level scope I would like to work on this :D

Anber: Ok, let's merge it! :)

Issue #562, opened 2020-03-31
Provide more global variables to node.vm context


jayu:
## Describe the feature

We should pass more variables to the global scope of the VM context. Eg. there is no `setTimeout` available in `VM` context. We need to investigate the common set of variables available in both node and browser and add them

We should spread the following to top-level scope of VM 

```
clearImmediate,
clearInterval,
clearTimeout,
setImmediate,
setInterval,
setTimeout
```

https://github.com/callstack/linaria/blob/672b0bfcedfba1c2a5a5e6f5bfae2d63bfeee241/src/babel/module.ts#L289-L297

## Motivation

More flexible and error-proof evaluation, see #399 `setTiemout` issue.

jayu:
closed via #592

PR #592, opened 2020-04-30
Added more variables when vm context is created

TMaszko:
## Motivation
Some variables available in global scope are not included by default in vm context See #562 

## Test plan
green CI


TMaszko:
Yeah you are absolutely right thanks 👍

jayu:
I appreciate the tests!

Issue #590, opened 2020-04-28
Add documentation on how to integrate Linaria


jayu:
## Describe the enhancement

Firstly we add more chapters to `docs/CONFIGURATION.md` to group working integrations
- [ ] Next.js heading and ling to [`with-linaria` example](https://github.com/zeit/next.js/tree/master/examples/with-linaria)
- [ ] Gatsby.js heading with a link to [gatsby-plugin-linaria](https://github.com/silvenon/gatsby-plugin-linaria) and optional custom config.
- [ ] CRA craco plugin https://github.com/jedmao/craco-linaria 
- [ ] Preact - based on config posted in https://github.com/callstack/linaria/issues/434#issuecomment-596817660
- [ ] Svelte - based on configs posted in https://github.com/callstack/linaria/issues/435#issuecomment-611226593

Then modify [Setup section in README](README.md#setup) and list there all possible working integrations from `docs/CONFIGURATION.md` with links obviously. 

I'm not sure what to do with Linaria [parcel plugin](https://github.com/callstack/parcel-plugin-linaria). I mean we can link it anyway, but I don't know if it is stable. 

## Motivation

There are tool/frameworks that work with linaria but integrations are not documented

## Related Issues

#434 , #435

amankkg:
What about [Snowpack](https://www.snowpack.dev/) integration?

There is a statement in their docs which seems to be false in case of Linaria:
> Snowpack also supports any popular CSS-in-JS library

arendjr:
@amankkg I made an attempt to integrate Linaria with Snowpack, using the following plugin:

```js
const babel = require("@babel/core");
const fs = require("fs");
const path = require("path");
const { promisify } = require("util");
const transform = require("linaria/lib/transform");

const readFile = promisify(fs.readFile);

module.exports = function (snowpackConfig, pluginOptions) {
  const { sourceMap, preprocessor, ...rest } = pluginOptions;

  return {
    name: "snowpack-linaria-plugin",
    resolve: {
      input: [".js", ".jsx", ".ts", ".tsx", ".mjs"],
      output: [".js", ".css"],
    },
    async load({ filePath }) {
      const contents = await readFile(filePath, "utf-8");
      const result = transform(contents, {
        filename: path.basename(filePath),
        preprocessor,
        pluginOptions: rest,
      });

      if (!result.cssText) {
        return {
          ".js": await babel.transformAsync(contents, { filename: filePath }),
        };
      }

      return {
[...]

PR #593, opened 2020-05-01
Document how to use Linaria with Svelte

### LINKED ISSUE #435: Support Svelte.
Currently compiler breaks when it encounters <script> tag:

```
ERROR in ./src/login/base/base.svelte
Module build failed (from ./node_modules/linaria/loader.js):
SyntaxError: /home/tony/Projects/auxilia/src/login/base/base.svelte: Unexpected token (1:0)

> 1 | <script>
    | ^
  2 |     import style from './styles';
  3 |     import Field from '../input/input.svelte';
  4 |     import { css } from 'linaria';
```
I suppose Vue would have very similar problem as compiler doesn't recognize that javascript is contained within the <script> tag.

Now for those who might not be aware, Svelte is a compiler for web standard compliant vanilla javascript components. Which made me think that Linaria sould be such a perfect fit to go hand in hand.

It's important to note that it actually does work, as I import generated class name, as the styles get compiled from separate file, but that beats the point of single-file-components.

madhavarshney:
## Motivation
Document how to use Linaria with Svelte. Resolves #435. 

## Summary
N/A

## Test plan
N/A


docs/BUNDLERS_INTEGRATION.md
```diff
@@ -248,3 +253,87 @@ export default {
...
+
+### Svelte
+
+#### Contents
+- [Svelte with Rollup](#Rollup)
+- [Svelte with Webpack](#Webpack)
+
+#### Rollup
+
+Take a look: [d964432](https://github.com/madhavarshney/svelte-linaria-sample/commit/d96443218694c0c8d80edf7c40a8fbf7c1f6997f)
```
madhavarshney, line 265: Would you like to create a separate examples folder / repo and instead link this there?

madhavarshney, line 265: Would you like to create a separate examples folder / repo and instead link this there? This may also help for #590.

jayu: That's a great idea! If we have docs on integrations in one place we can also have examples in our repo, especially if the example does not exist in specific tools repository (eg. next.js has an example with linaria). 

Would it be possible to have one example for svelte but with two bundlers set-up? 
If not I'm ok with the following structure
`/examples/svelte/rollup`
`/examples/svelte/webpack`

But for any other examples. I believe there will be possible to have one code base with two bundlers or there will be only one bundler if framework assumes usage of a specific one

jayu:
@madhavarshney I'm merging it as it is. If you decide to add svelte example just open another PR. Thanks for the input!

PR #594, opened 2020-05-01
fix: handle blank string values

chrisabrams:
This will handle blank string values. Sometime after version 1.3.3 the library started throwing an error about being unable to compile blank string values. Now it will just skip the value instead of throwing an error.

## Motivation

We have some mixins which will return a blank string if no conditions are met. Our expectation is that a blank string will cause the CSS property to be ignored (not included) by the compiler.

## Summary

This PR will cause Linaria to skip blank string values, aka '', instead of throwing an error that linaria does not know how to compile the value.

## Test plan

`yarn test`

### DIFF
--- a/src/__tests__/__snapshots__/babel.test.ts.snap
+++ b/src/__tests__/__snapshots__/babel.test.ts.snap
@@ -19,6 +19,24 @@ const title = String.raw\`This is something\`;"
 
 exports[`does not output CSS if none present 2`] = `Object {}`;
 
+exports[`does not output CSS property when value is a blank string 1`] = `
+"import { css } from 'linaria';
+export const title = \\"th6xni0\\";"
+`;
+
+exports[`does not output CSS property when value is a blank string 2`] = `
+
+CSS:
+
+.th6xni0 {
+  font-size: ;
+  margin: 6px;
+}
+
+Dependencies: NA
+
+`;
+
 exports[`evaluates and inlines expressions in scope 1`] = `
 "import { styled } from 'linaria/react';
 const color = 'blue';
--- a/src/babel/evaluators/templateProcessor.ts
+++ b/src/babel/evaluators/templateProcessor.ts
@@ -146,6 +146,11 @@ export default function getTemplateProcessor(options: StrictOptions) {
             const value = valueCache.get(ex.node);
             throwIfInvalid(value, ex);
 
+            // Skip the blank string instead of throwing an error
+            if (value === '') {
+              return;
+            }
+
             if (value && typeof value !== 'function') {
               // Only insert text for non functions
               // We don't touch functions because they'll be interpolated at runtime
--- a/src/__tests__/babel.test.ts
+++ b/src/__tests__/babel.test.ts
@@ -342,6 +342,22 @@ it('does not output CSS if none present', async () => {
   expect(metadata).toMatchSnapshot();
 });
 
+it('does not output CSS property when value is a blank string', async () => {
+  const { code, metadata } = await transpile(
+    dedent`
+    import { css } from 'linaria';
+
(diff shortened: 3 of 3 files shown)


src/babel/evaluators/templateProcessor.ts
```diff
@@ -146,6 +146,11 @@ export default function getTemplateProcessor(options: StrictOptions) {
             const value = valueCache.get(ex.node);
             throwIfInvalid(value, ex);
 
+            // Skip the blank string instead of throwing an error
+            if (value === '') {
+              return;
```
jayu, line 151: I was wondering about introducing a warning here, but I think we can leave tracking of potential errors to users. In your case, it will probably introduce a ton of unwanted warnings. 
It is logically correct to skip the CSS property if the value is empty.

src/__tests__/__snapshots__/babel.test.ts.snap
```diff
@@ -19,6 +19,24 @@ const title = String.raw\`This is something\`;"
...
+"import { css } from 'linaria';
+export const title = \\"th6xni0\\";"
+`;
+
+exports[`does not output CSS property when value is a blank string 2`] = `
+
+CSS:
+
+.th6xni0 {
+  font-size: ;
```
jayu, line 32: It seems to not work 🤔 Strange I don't know why it happens. If you will not be able to solve this I can debug the issue, let me know

src/babel/evaluators/templateProcessor.ts
chrisabrams: Skipping the CSS property if the value is "falsy" has been the default for CSS processors for a long time.

Here are some examples:
- [SASS](https://markdotto.com/2019/08/30/null-sass-variables/)
- [CSS Nano](https://cssnano.co/optimisations/)

src/__tests__/__snapshots__/babel.test.ts.snap
chrisabrams: I only tested ES6 template interpolations; I didn't test a literal blank value like that. I would expect the CSS processor to pick that up though. When you say it doesn't work; what does happen?

chrisabrams:
@jayu its starting to look like a library design decision here. Do you want linaria to act like default CSS processors and treat falsy / blank values the same way? Do you want that to be a default? I would argue that an optional option could be to throw an error on falsy values instead of skipping them in the compilation. What do you think? I could expand this PR to address the topic if so.

src/__tests__/__snapshots__/babel.test.ts.snap
jayu: tl;dr: behavior is correct 

Simply there should not be `font-size: ;` line in the snapshot if we `return` and do not process. But maybe it is because of how snapshots are prepared, I don't know honestly.

Oh I know! It will not remove the `font-size: ;` because we iterate over quasis and append this string here https://github.com/callstack/linaria/blob/d596c0fffe97f8de3af0c423adbd984e8497012e/src/babel/evaluators/templateProcessor.ts#L103
And then we get matching expression that follows processed quasi, and then we skip adding it's value. So the overall result is the same because we leave empty string there 😅 (I recommend checking how babel parse the target templates expressions on https://astexplorer.net/ if you are interested)

I think we can leave it as it is because otherwise, we would have to, do additional work. In the CSS output, the line is removed, probably by `stylus` preprocessor that we use in `transform.ts`https://github.com/callstack/linaria/blob/master/src/transform.ts#L145

jayu:
@chrisabrams I think I answered your question in my previous comment. Thanks for the note about the preprocessors! I merging ;)

chrisabrams:
@jayu thanks!

Issue #434, opened 2019-06-01
Linaria with preact-cli & preact X


secretlifeof:
Does anybody have experience with setting up Linaria with Preact-cli/Preact X?

I tried the following:
linaria moduleWrapper
```
"moduleNameMapper": {
			"\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/tests/__mocks__/fileMock.js",
			"\\.(css|less|scss)$": "identity-obj-proxy",
			"^./style$": "identity-obj-proxy",
			"^preact$": "<rootDir>/node_modules/preact/dist/preact.min.js",
			"^react$": "preact/compat",
			"^react-dom$": "preact/compat",
			"^linaria$": "preact/compat",
			"^create-react-class$": "preact-compat/lib/create-react-class",
			"^react-addons-css-transition-group$": "preact-css-transition-group"
		}
```
.babelrc
```
{
  "env": {
    "test": {
      "presets": [
        ["preact-cli/babel", { "modules": false }],
        "linaria/babel"
      ]
    }
  }
}
```
preact.config.js
```
const MiniCssExtractPlugin = require('mini-css-extract-plugin');

export default {
	webpack(config, env, helpers, options) {
		// config.resolve.alias = Object.assign({}, config.resolve.alias, {
		// 	react$: 'preact/compat/',
		// 	'react-dom$': 'preact/compat',
		// 	react: 'preact/compat/',
		// 	'react-dom': 'preact/compat',
		// 	linaria: 'preact/compat'
		// });

		config.plugins.push(new MiniCssExtractPlugin({ filename: 'styles.css' }));

		config.module.rules.push({
			test: /\.m?js$/,
			exclude: /(node_modules)/,
			use: [
				{
					loader: 'babel-loader'
				},
				{
					loader: 'linaria/loader'
				}
			]
		});

		return config;
	}
};
```

The following error is returned:
**bundle.esm.js:1593 Error: Using the "styled" tag in runtime is not supported. Make sure you have set up the Babel plugin correctly. Is .babelrc not setup correctly?**

Versions:
"linaria": "^1.3.1"
"preact": "^10.0.0-beta.2"
"preact-compat": "^3.17.0"
"preact-render-to-string": "^5.0.3"

imbsky:
I use linaria with preact.
https://github.com/imbsky/bsky-moe

[...]

PR #595, opened 2020-05-02
Docs/preact config

Tsubasa1218:
## Motivation

Add steps to configure Linaria when using Preact/Preact CLI.

This would help #434 , and #590 

## Summary

- Added configuration for preact in `docs/CONFIGURATION.md`,
- Added a new bullet point in README.md for integration links.
- Added an example since it was requested in #593 
- Because I added some code from preact in the examples folder, I added a folder int the exludes in `.eslintignore` file

## Test plan

N/A


examples/Preact/preact-example/src/components/app.js
```diff
@@ -0,0 +1,32 @@
...
+	 *	@param {Object} event		"change" event from [preact-router](http://git.io/preact-router)
+	 *	@param {string} event.url	The newly routed URL
+	 */
+	handleRoute = e => {
+		this.currentUrl = e.url;
+	};
+
+	render() {
+		return (
+			<div id="app">
```
jayu, line 22: Since the usage of linaria is in Header I would suggest to put it also in this file.  Most of the users will look for example usage in the main app file. It can be just 
```js
const Container = styled.div`
background-color: 'lightblue'
`
```

Tsubasa1218: Yeah, you are right xD

I added one example with `css` and the other with `styled`

jayu: Awesome, thanks!

PR #598, opened 2020-05-04
ESM

wereHamster:
## Motivation

Make the package available as ES Modules.

## Summary

 1. Build a second set of dist files which use ESM syntax (export/import), into the esm/ folder.
 2. Make these modules available through "module" package.json field.

## Notes

I removed both react.js and react.d.ts from the package root, and instead created react/package.json. This is required so that bundlers can access the correct entry point (main / module). The same could be done with other top-level modules (server.js, rollup.js, loader.js, …). The next version of Node.js (v13+) will provide a [much better method](https://nodejs.org/api/esm.html#esm_package_entry_points) for that, but that will take some time until the bundlers add support for it.

Here is an example with the new Node.js package entry points: https://github.com/callstack/linaria/compare/wereHamster:esm...wereHamster:package-entry-points


react/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
jayu, line 1: Is this "partial" package.json supported officially? I have never seen such syntax before

wereHamster: Should be. Node.js looks for the closest package.json and uses it. `@material-ui/core` does the same. See here: https://unpkg.com/browse/@material-ui/core@4.9.13/Avatar/package.json

jayu: Thanks, I will check it

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "2.0.0-alpha.2",
   "description": "Blazing fast zero-runtime CSS in JS library",
   "main": "lib/index.js",
+  "module": "esm/index.js",
   "bin": "bin/linaria.js",
   "files": [
```
jayu, line 8: please replace `react.js` to `react` in files array

jayu:
Do I understand the changes regarding the building correctly? 

You added two environments `legacy` and `test`.
Both newly added environments have the same setup.
We use the environment `legacy` to build `lib/`
We do not use any environment to build `esm/`, hence `preset-env` is not used and modules are not transpired and we get `import/export` statements

What is the `test` environment used for?

wereHamster:
I believe jest sets NODE_ENV=test implicitly, but I can quickly test without the test env in babel.config.js to see if it breaks anything…

wereHamster:
Yup, without the test env jest fails, because jest doesn't understand ES modules.

<details>

```
● replaces dynamic imports with a noop

    …/linaria/src/babel/dynamic-import-noop.ts:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import syntax from '@
babel/plugin-syntax-dynamic-import';
                                                                                                    ^^^^^^

    SyntaxError: Unexpected identifier
```

</details>

jayu:
Got it, so can you please refactor the config to not have repeated parts? And put a note there, that how the config works, meaning if we don't use any env variable it will produce `ESM`, `legacy` is used to build common.js, and since we changed the config and there is no `default` config, we need to set the additional config for test environment to make tests work.

jayu:
@wereHamster I had to fix the core-js test, I opened a PR to your fork https://github.com/wereHamster/linaria/pull/1

wereHamster:
@jayu sorry about that, I saw that all checks are green and didn't know that there were any tests that are not being run on circleci.

jayu:
@wereHamster No problem, I implemented  this test incorrectly, it didn't fail while the needed environment variable was not provided. I didn't want you to have to refactor this pr once again so I did it myself :D I will take care of merging tomorrow probably

wereHamster:
@jayu could you release this to npm so I can test it in my project? Thanks.

jayu:
@wereHamster will do on Monday. I tried at friday, but npm was down then.

Issue #597, opened 2020-05-03
Regression in styles evaluation in `1.4.x` and above


wereHamster:
## Environment

- Linaria version: v2.0.0-alpha.2
- Bundler (+ version): rollup
- Node.js version: v10
- OS: macOS

## Description

In v1 code like this works. In v2 linaria fails with the error `classes is not defined.`:

```
const classes = {
    cell: css`
      opacity: 0;
    `,
  };

  const className = css`
  &:hover .${classes.cell}) {
    opacity: .2;
  }
  `;
```

## Reproducible Demo

https://github.com/wereHamster/linaria-v2-regression/blob/master/input.js

jayu:
@wereHamster Which version of Linaria you used before the update to `2.0`? I observed that this syntax didn't work before, but I don't remember in which version.  

Btw. You should avoid using linaria in not top-level scope, if you move quotes code up, it will work.

wereHamster:
v1.3.3.

What do you mean not in top-level scope? I've had a lot of code that is something like that:

```
function Component() {
  return <div className={css`<plain CSS here>`} />
}
```

Is that also not allowed? Do I need to pull all those css`` blocks out of the component into top-level scope?

(For reference, here is one such example: https://github.com/timvir/timvir/blob/ba2ee527466a5ab503ee61988171f8203c10a56c/src/components/ColorContrastInspector/internal/Cell.tsx)

jayu:
@wereHamster Recently we've been thinking if we can restrict using linaria just to top-level scope. Some people try to use Linaria like this :`

```
function Component(props) {
  return <div className={css`color: ${props.color}`} />
}
```
Which will no work, because in built time we don't know what is the value of `props.colors`.
We've been thinking about restricting, but we didn't do it yet, because there might be many users that will have to refactor the codebase.  

According to current Linaria assumptions, this code works

```function Component() {
  return <div className={css`<plain CSS here>`} />
}
```
and this code should also work, but it doesn't (which is regression as you mentioned, not only in `2.0` but also in `1.4` probably)

```function Component() {
  const color = 'red'
  return <div className={css`color: ${color}`} />
}
```
So for now you don't have to pull everything into top-level scope, but we think that is generally a great idea to have styles defined even in separate file, or at least in top-level scope to make code better organized.

jayu:
@wereHamster I adjusted the Issue name, I hope you don't mind :)

jayu:
[...]

PR #604, opened 2020-05-05
Fix regression with `css` in nested scopes (#597)

Anber:
## Summary

This PR fixes issue #597

## Test plan

New tests were added


src/babel/utils/isStyledOrCss.ts
```diff
@@ -0,0 +1,64 @@
+import { types as t } from '@babel/core';
+import { NodePath } from '@babel/traverse';
+import hasImport from './hasImport';
+import { State, TemplateExpression } from '../types';
+
+type Result = NonNullable<TemplateExpression['styled']> | 'css' | null;
+
+const cache = new WeakMap<NodePath<t.TaggedTemplateExpression>, Result>();
```
jayu, line 8: Do I understand correctly and this cache will be created for each file that that is going to be transformed by `transform.ts`? Node caches the required modules, and this module should be required every time linaria/babel is used?

src/babel/evaluators/templateProcessor.ts
```diff
@@ -33,12 +33,21 @@ function hasMeta(value: any): value is StyledMeta {
...
+const cache = new WeakSet();
+
 export default function getTemplateProcessor(options: StrictOptions) {
   return function process(
     { styled, path }: TemplateExpression,
     state: State,
     valueCache: ValueCache
   ) {
+    if (cache.has(path)) {
+      // Do not process one node twice
```
jayu, line 45: Why it happed?

src/babel/utils/isStyledOrCss.ts
Anber: Yep. It's a lightweight local cache for avoiding some repetitive operations. Since we use `WeakMap` we don't need to control memory leaks and cache invalidation here.

src/babel/evaluators/templateProcessor.ts
Anber: It happens if one `css` tag is referenced in a template more than once. After the first pass, it will be replaced with `StringLiteral` and the second pass will fail because it expects `TaggedTemplateExpression`.

satya164: Maybe we can add this comment into the file :)

PR #605, opened 2020-05-06
Remove TypeScript module augmentation of @babel/core

### LINKED ISSUE #599: @types/babel__core is an unlisted dependency
## Environment

- Linaria version: v2.0.0-alpha.2
- Bundler (+ version): babel
- Node.js version: 13
- OS: macOS

## Description

I received the following error from TypeScript:

```
./node_modules/.bin/tsc --build src/tsconfig.json
node_modules/linaria/lib/babel/types.d.ts:94:16 - error TS2665: Invalid module name in augmentation. Module '@babel/core' resolves to an untyped module at '<project>/node_modules/@babel/core/lib/index.js', which cannot be augmented.

94 declare module '@babel/core' {
                  ~~~~~~~~~~~~~
```

The app works fine, it's just the type checke that's not happy. I was able to fix this by installing `@types/babel__core` myself, though it's probably not nice to force users of this library to manually install this dependency.

## Reproducible Demo

https://github.com/timvir/timvir/pull/124

```
npm ci
./node_modules/.bin/tsc --build src/tsconfig.json
```

wereHamster:
Fixes #599

### DIFF
--- a/src/babel/evaluators/shaker/GraphBuilderState.ts
+++ b/src/babel/evaluators/shaker/GraphBuilderState.ts
@@ -1,6 +1,7 @@
 import { types as t } from '@babel/core';
 import ScopeManager from './scope';
 import DepsGraph from './DepsGraph';
+import { VisitorKeys } from '../../types';
 
 export type OnVisitCallback = (n: t.Node) => void;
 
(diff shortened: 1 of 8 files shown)


jayu:
How can I check if this actually fixes the issue? We still use types from babel core, I think it will not help much. Our types still relay on `@types/babel__core` so it has to be a dependency of Linaria :/

wereHamster:
Closing in favour of #607

PR #606, opened 2020-05-07
patch: issue 597

jayu:
## Motivation

Patch to PR with statements hoisting that fixes two issues:
- sometimes it happened that statements were in different orders which result in the import was defined below its usage.
- during hoisting, if an expression was an identifier, it was replaced with a new identifier that refers to a hoisted value. New identifiers didn't have `start`, `end`, `loc` which result in an error in `templateProcessor`.

## Summary

I ported sorting statements from `1.3.x`

I make Linaria use a clone of the original statement during evaluation.

Additionally, I added tests for the usage of `styled` in not top-level scope similarly to `css`

## Test plan
New tests added

### DIFF
--- a/src/babel/visitors/CollectDependencies.ts
+++ b/src/babel/visitors/CollectDependencies.ts
@@ -15,7 +15,9 @@ import isStyledOrCss from '../utils/isStyledOrCss';
  */
 function hoist(ex: NodePath<t.Expression | null>) {
   const Identifier = (idPath: NodePath<t.Identifier>) => {
-    if (!idPath.isReferencedIdentifier()) return;
+    if (!idPath.isReferencedIdentifier()) {
+      return;
+    }
     const binding = idPath.scope.getBinding(idPath.node.name);
     if (!binding) return;
     const { scope, path: bindingPath, referencePaths } = binding;
@@ -29,9 +31,10 @@ function hoist(ex: NodePath<t.Expression | null>) {
       hoist(initPath);
       initPath.hoist(scope);
       if (initPath.isIdentifier()) {
-        referencePaths.forEach(referencePath =>
-          referencePath.replaceWith(t.identifier(initPath.node.name))
-        );
+        referencePaths.forEach(referencePath => {
+          referencePath.replaceWith(t.identifier(initPath.node.name));
+          // referencePath.node.name = initPath.node.name;
+        });
       }
     }
   };
@@ -54,7 +57,6 @@ export default function CollectDependencies(
   if (!styledOrCss) {
     return;
   }
-
   const expressions = path.get('quasi').get('expressions');
 
   debug('template-parse:identify-expressions', expressions.length);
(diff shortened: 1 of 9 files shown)


src/babel/visitors/CollectDependencies.ts
```diff
@@ -29,9 +31,10 @@ function hoist(ex: NodePath<t.Expression | null>) {
       hoist(initPath);
       initPath.hoist(scope);
       if (initPath.isIdentifier()) {
-        referencePaths.forEach(referencePath =>
-          referencePath.replaceWith(t.identifier(initPath.node.name))
-        );
+        referencePaths.forEach(referencePath => {
+          referencePath.replaceWith(t.identifier(initPath.node.name));
+          // referencePath.node.name = initPath.node.name;
```
Anber, line 36: del

src/babel/visitors/CollectDependencies.ts
```diff
@@ -66,13 +68,22 @@ export default function CollectDependencies(
...
+        const originalExNode = t.cloneNode(ex.node);
+
         hoist(ex);
-        return { kind: ValueType.LAZY, ex };
+
+        // save hoisted expression to be used to evaluation
+        const hoistedExNode = t.cloneNode(ex.node);
+
+        // get back original expression to the tree
+        ex.replaceWith(originalExNode);
```
Anber, line 84: I don't get it. Why you replace hoisted node by the original one?

Anber: Is it possible to just store that original location data instead of cloning node?

jayu:
@Anber will do tomorrow

src/babel/visitors/CollectDependencies.ts
jayu: The quick solution for this problem was to rename the identifier instead of replacing it with a new node. This is the commented line
`referencePath.node.name = initPath.node.name;`

The other solution would be probably to replace a node but assign position of the old node to it. However, I'm not sure if the position will be current since the length of the name of the new identifier would be different than the original one.  I'm not sure if this will affect further processing, but the location is later used in `templateProcessor` so it has to be defined and would be good if it would be accurate. 

The other thing about this is error message. On the evaluation fails for something like this
```
css`
    font-size: ${valA + valB}
`
```
you will probably see the error frame with

```
css`
    font-size: ${_ref + _ref2}
`
```
Which is a bad DX. It is a rare case, but possible. Also, it is more informative while debugging to have printed the original and transformed statement.

That's why I decided to use 'artifical' node for evaluation, and preserve the original node in place where it was used. I don't feel good about the complexity that this solution introduces, but I would feel safer to have this, instead of changed identifier name or assigned incorrect location to new identifier.

Anber: Sounds reasonable. Let's merge it, and maybe later we will find a better solution.

PR #609, opened 2020-05-11
fix: use transform-tagged-template-literal plugin explicitly

jayu:
## Motivation

**Update:** it closes #609 

I initially wanted to refactor the naming of some files and add more documentation headers, but I found this strange bug. 
It happened that we relay on `@babel/plugin-transform-template-literal` during the evaluation of dependencies (preval). It happened because we had `preset-env` in Linaria babel config which includes this plugin, and most of the users probably also used this preset, so no one had the problem. 

But once the plugin is excluded, preval stops working because we don't collect all styled-components. Here is an explanation that I leave in a comment: 
```
   We made a mistake somewhen, and linaria preval was dependent on `plugin-transform-template-literals`
Usually it was loaded into preval, because user was using `@babel/preset-env` which included that plugin. Internally we used this preset for tests (and previously for everything)
The ordering is very important here, that's why it is added as a preset, not just as a plugin. It makes this plugin run *AFTER* linaria preset
In preval we have 2 visitors, one for Call Expressions and second for TaggedTemplateLiterals. Babel process TaggedTemapltes first, and we grab css`` statements.
Then it process TaggedTemplate with mentioned plugin, which transforms it to CallExpression.
Then Linaria can identify all `styled` as call expressions.
Presets ordering is from last to first, so we add it at the beginning of the list.
note that in shaker, preset-env is used explicitly
```

## Summary

- moved visitors used only in preval to `extractors/visitors`
- renamed visitors to better reflect their purpose
- added documentation headers
- added usage of mentioned plugin explicitly as a preset to persist needed ordering
- removed `preset-env` from our test config which makes config much more appropriate

## Test plan

I removed `preset-env` from babel config for tests, but all tests should still work since we added plugin explicitly. 

## Thoughts

- We could be better and refactor code to not have to rely on this plugin at all, but I treat this PR as a bug patch, we can improve it in the future. 
- I treat moving visitors to a subdirectory as the beginning of redefining the file structure. It should better reflect the separate stages of Linaria processing to make it easier to understand the code. 
- We should redefine current nomenclature like `evaluation`, `preval` etc. since maintainers know what it means, but newcomers may be confused. I will try to propose something in this week


src/babel/evaluators/extractor/index.ts
```diff
@@ -51,6 +51,18 @@ const extractor: Evaluator = (filename, options, text, only = null) => {
     { useESModules: false },
   ]);
 
+  // We made a mistake somewhen, and linaria preval was dependent on `plugin-transform-template-literals`
+  // Usually it was loaded into preval, because user was using `@babel/preset-env` which included that plugin. Internally we used this preset for tests (and previously for everything)
+  // The ordering is very important here, that's why it is added as a preset, not just as a plugin. It makes this plugin run *AFTER* linaria preset
+  // In preval we have 2 visitors, one for Call Expressions and second for TaggedTemplateLiterals. Babel process TaggedTemapltes first, and we grab css`` statements.
```
TMaszko, line 57: TaggedTemplates ;)

TMaszko: I don't really understand `Babel process TaggedTemapltes first, and we grab css`` statements.` this statement because I don't know whether it describes past situation or current. This should be said more explicitly

src/babel/evaluators/visitors/ProcessCSS.ts
```diff
@@ -1,10 +1,13 @@
+/**
+ * This visitor replaces css tag with the generated className
+ *
+ */
+
 import { NodePath } from '@babel/traverse';
 import { types as t } from '@babel/core';
-import getLinariaComment from '../utils/getLinariaComment';
+import getLinariaComment from '../../utils/getLinariaComment';
```
TMaszko, line 8: We may think about adding aliases in the future to avoid long relative paths

jayu: I like aliases but it is hell in non-app projects :(

src/babel/evaluators/extractor/index.ts
jayu: thanks, I will adjust this comment to not be a mind shortcut :P

src/babel/evaluators/visitors/ProcessCSS.ts
TMaszko: I trust you :D but maybe nowadays It is easier than before.. anyway It is worth to consider it in the future :)

jayu: We will refactor everything to monorepo in interest future so it will most likely improve the situation

Issue #244, opened 2018-10-10
`fragment` tag for composition


satya164:
## Problem

Currently it's not possible to compose styles from multiple classnames together. Previously we had a `include` helper which would take a class name and return the unprocessed CSS text. I removed it because it wasn't strightforward to implement it with the new approach and I haven't needed to use it personally. In addition, an extra class name was generated whether you used it anywhere or not.

Another usecase is that once you get into an interpolation inside the tagged template, we don't parse the nested interpolations. So it's not possible to do something like following:

```js
const Box = styled.div`
  font-size: 14px;

  ${borders
    .map(({ p, w }) => `border-${p}: ${w}px solid ${props => props.color}`)
    .join('\n')};
`;
```

`styled-components` has exports `css` tag to support this. However, the `css` tag is used for class names in Linaria, so we can't do the same.

## Prior art

Emotion allows you to interpolate a class name which inlines its styles: https://emotion.sh/docs/composition

One problem with this approach is that you're interpolating a class name, but getting the CSS text instead. This breaks the following use case where you might want to refer to another class name:

```js
const paragraph = css`
  font-size: 14px;
`;

const article = css`
  backgrounc-color: white;

  .${paragraph} {
    font-size: 16px;
  }
`;
```

## Current solution

Currently, you could do something like this:

```js
const shared = `
  font-size: 18px;
`;

const Title = styled.h1`
  ${shared};

  color: black;
`;
```

There are 2 issues with this approach:

1. There's no syntax highlighting or autocomplete for the CSS
2. It's not possibe to use dynamic interpolations like you would in `styled`

## Proposal

We could add a new `fragment` tag to allow this use case:

```js
const shared = fragment.css`
  font-size: 18px;
`;

const Title = styled.h1`
  ${shared};

  color: black;
`;
```

The `fragment` tag itself won't generate any class names, but when the result is interpolated, it will inline the CSS text along with the dynamic interpolations. It'll work very similarly to how `styled-components` handles interpolation of `css` tagged template literals. As for parsing, it would be parsed similar to how the `styled.x` tags are parsed.

*Note: Initially I thought of having just `fragment`, but went for `fragment.css` since it allows us to have syntax highlighting and autocomplete.*

Since the `fragment` tag won't generate a class name, to use it as a class, you could do something like this:

```js
const title = css`${shared}`;
```

You could even inline it if you don't like creating an extra variable:

```js
<h1 class={css`${shared}`} />
```

## Why not add it

[...]

PR #615, opened 2020-05-19
feat: inlined styles composition

jayu:
## Motivation

#244 
And being able to share styles between web and native in future RN support

## Summary

Todo:
- [ ] fix existing tests
- [ ] add new tests
- [ ] optimize if possible

Features:
- can compose `css` with other `css` or `styled`
- styles are composed by inlining, so place of the composition affects styling (intentional behaviour)
- `css` can contain dynamic interpolation, it has no effect once used as a class, it has effect once interpolated in styled component
- composition inside sub-selectors is possible
- composition of multiple tags together is possible

the following code is an example of valid usage

```js
const fragment = css`
  color: red;
  font-size: 40px;
  padding-top: ${(props) => props.height || '200px'}; 
`

const anotherFragment = css`
  color: yellow;
  ${fragment};
  border: 10px solid black;
`

const component styled.h1`
color: black;
${anotherFragment}
.sub-class {
    ${fragment}
}
`

```

Cons:
Increases build time by 15% comparing to `2.0.0-alpha.5` (with shaker as default)

implementation details: 
TBD

## Test plan

Green CI, new tests added

### DIFF
--- a/src/babel/visitors/CollectDependencies.ts
+++ b/src/babel/visitors/CollectDependencies.ts
@@ -9,7 +9,6 @@ import throwIfInvalid from '../utils/throwIfInvalid';
 import { State, StrictOptions, ValueType, ExpressionValue } from '../types';
 import { debug } from '../utils/logger';
 import isStyledOrCss from '../utils/isStyledOrCss';
-
 /**
  * Hoist the node and its dependencies to the highest scope possible
  */
@@ -64,7 +63,8 @@ export default function CollectDependencies(
   const expressionValues: ExpressionValue[] = expressions.map(
     (ex: NodePath<t.Expression>) => {
       const result = ex.evaluate();
-      if (result.confident) {
+      // eslint-disable-next-line
+      if (false && result.confident) {
         throwIfInvalid(result.value, ex);
         return { kind: ValueType.VALUE, value: result.value };
       }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,3 @@
 {
-  "typescript.tsdk": "website/node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
--- a/src/__tests__/__snapshots__/babel.test.ts.snap
+++ b/src/__tests__/__snapshots__/babel.test.ts.snap
@@ -42,7 +42,11 @@ exports[`evaluates and inlines expressions in scope 1`] = `
 const color = 'blue';
 export const Title = /*#__PURE__*/styled(\\"h1\\")({
   name: \\"Title\\",
-  class: \\"th6xni0\\"
+  class: \\"th6xni0\\",
+  vars: {
+    \\"th6xni0-0\\": [color],
+    \\"th6xni0-1\\": [100 / 3, \\"%\\"]
+  }
 });"
 `;
 
@@ -51,8 +55,8 @@ exports[`evaluates and inlines expressions in scope 2`] = `
 CSS:
 
 .th6xni0 {
-  color: blue;
-  width: 33.333333333333336%;
+  color: var(--th6xni0-0);
+  width: var(--th6xni0-1);
 }
 
 Dependencies: NA
@@ -228,7 +232,10 @@ const styles = [{
 }];
 export const Title = /*#__PURE__*/styled(\\"h1\\")({
   name: \\"Title\\",
-  class: \\"th6xni0\\"
+  class: \\"th6xni0\\",
+  vars: {
+    \\"th6xni0-0\\": [styles]
+  }
 });"
 `;
 
@@ -237,7 +244,7 @@ exports[`inlines array styles as CSS string 2`] = `
 CSS:
 
 .th6xni0 {
-  flex: 1; display: block; height: 24px;
+  var(--th6xni0-0)
 }
 
 Dependencies: NA
@@ -274,7 +281,10 @@ const cover = {
 };
 export const Title = /*#__PURE__*/styled(\\"h1\\")({
   name: \\"Title\\",
-  class: \\"th6xni0\\"
+  class: \\"th6xni0\\",
+  vars: {
+    \\"th6xni0-0\\": [cover]
+  }
 });"
 `;
 
@@ -283,7 +293,7 @@ exports[`inlines object styles as CSS string 2`] = `
 CSS:
 
 .th6xni0 {
-  position: absolute; top: 0; right: 0; bottom: 0; left: 0; opacity: 1; min-height: 420px; &.shouldNotBeChanged { border-color: #fff; } @media (min-width: 200px) { -webkit-opacity: 0.8; -moz-opacity: 0.8; -ms-opacity: 0.8; -o-opacity: 0.8; -webkit-border-radius: 2px; -moz-border-radius: 2px; -ms-border-radius: 2px; -o-border-radius: 2px; -webkit-transition: 400ms; -moz-transition: 400ms; -o-transition: 400ms; -ms-transition: 400ms; }
+  var(--th6xni0-0)
 }
 
 Dependencies: NA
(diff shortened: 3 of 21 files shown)


src/babel/visitors/CollectDependencies.ts
```diff
@@ -64,7 +63,8 @@ export default function CollectDependencies(
   const expressionValues: ExpressionValue[] = expressions.map(
     (ex: NodePath<t.Expression>) => {
       const result = ex.evaluate();
-      if (result.confident) {
+      // eslint-disable-next-line
+      if (false && result.confident) {
```
TMaszko, line 67: Hmm :D It's always false :D

brandonkal:
Don't do this. Too much magic and very confusing. The value of the css function is a class name string.

There's are too many "ifs" here. CSS should be simple, this is a complex solution to a simple problem.

Your example component is red so you don't need all that code.

jayu:
This is just example usage, so I guess users can use this how they want 🤔

src/babel/visitors/CollectDependencies.ts
jayu: Thanks, I forgot to remove this :P

nterziyski:
@jayu This is definitely a feature we have been looking forward to, when considering Linaria. The way our and a lot of other projects, based on either styled-components or emotion, are structured has style composition ingrained. By having this out of the box, we could migrate a project in less than a week, rather than rewrite numerous components. 👍 Let me know if I can somehow help get this out faster!

brandonkal:
Of course it is just an example usage. But it doesn't demonstrate a good use case and only shows why this is a bad API. 

@nterziyski Why would you want css to have a double meaning? Please give an example where you are not better served by a modifier class or an exported fragment function. It is doing something here that it shouldn't. Static analysis suggests the value of "fragment" in the example is a classname. But "if" x then it is not.

nterziyski:
@brandonkal We use extensive style composition in our emotion-based project. From the top of my head:
```
const pointer = css`
  cursor: pointer;
`
const inset = css`
  padding: 8px;
`
const bottomSpacer = css`
  margin-bottom: 8px;
`
const trivialButtonLikeElement = css`
  // ...element specific styles like color, width, height etc.
  ${pointer};
  ${inset};
`
const anotherButtonLikeElement = css`
  // ...element specific styles like color, width, height etc.
  ${pointer};
  &.with-inset {
    ${inset};
  }
  ${bottomSpacer};
`
```
You can arguably achieve the same behaviour by using:
```className={classnames(anotherButtonLikeElement, pointer, inset)}```
but this covers only the most trivial composition.

From where I stand, @jayu 's implementation will streamline the migration process from other popular CSS-in-JS libraries to Linaria

**Comment by satya164** [comment]
I agree with @brandonkai that this can be confusing. Treating the same string as a class name or a style based on guessing how it's used, is problematic. It's better to avoid such heuristics.

It's also worth noting that linaria's babel plugin doesn't parse CSS, it doesn't know the actual usage. It can only guess what's the intention, but guesses can be wrong, and probably shouldn't be the primary API. I think heuristics are great for stuff like warnings though.

There's also the topic of it being confusing, if I can interpolate the string to get styles, I'd expect it to contain the CSS string which I would be able to interpolate somewhere else and get the same behaviour. Originally Linaria was intended to be a build-time optimization - you write code youd normally do in runtime and it transparently gets optimized during build time. If something isn't possible during run time, we wouldn't do it in build-time optimization step.

Another one is props interpolation inside `css` tag. What happens if the result gets used as a class name? It's better if APIs make invalid usage impossible.

First version of Linaria had an API called `compose`, where you could achieve similar behaviour:

```js
const shared = css`
  font-size: 18px;
`;

const title = css`
  ${compose(shared)}

  color: black;
`;
```

In this case, it was explicit that you're composing another class name, the class name itself didn't return the styles, but the `compose` function did by taking a class name. It was removed in current version because the way Linaria worked changed and it wasn't something straightforward to implement. It also couldn't interop with the new `styled` API that was added because interpolating props inside `css` is invalid.

The idea behind my proposal in #244 was to avoid ambiguity and have a clear API for composition. It's a new API since how it's used isn't same as how `css` is used. A fragment wouldn't return a class name, it's more akin to interpolating a string. The only reason it's even a different API is because of `styled` where we need special syntax for interpolating props.

@nterziyski regarding your use case, I agree that it'd be nice to be able to migrate such code, but there's an important difference here, `css` tag in Linaria returns a class name that can be used as selector, in Emotion, it doesn't (or at least it's deprecated afaik), so we can't expect them to be inter-operable without changing the core API. Probably it'd be better to write a codemod to do this instead of having heuristics in an API.

jayu:
@satya164  I do not understand the part about heuristics. Due to the current API the only way how class can be interpolated inside `css` or `styled` to be used as the actual class is to have it prepended with a dot `.` I don't see any other cases, maybe with exception for attribute the selector like `[class=]` but I have never seen such usage, to be honest. 
I don't know if you looked into implementation, but we actually do parse CSS in babel now :P 

The only part I don't like about introduced API is the confusion behind using props interpolation in `css`, but I believe this can be well explained in the docs. 

I should add a few words about of explanation of implementation but didn't have time for that.

My goal with this feature is to open a route to share styles between react and react-native with Linaria. As a side effect, it turned out that it helps to migrate projects to Linaria due to the fact that API is similar to `emotion` API.

satya164:
The problem is that this is overloading the meaning of `css` depending on if there's a dot in front of it. It makes the API not easily explainable. 
Emotion used to have such an API too, and they have now deprecated it.

> I don't see any other cases, maybe with exception for attribute the selector like [class=]

Maybe there is some utility function which takes class name and returns some styles which use that class name interpolated as a nested selector, maybe some utility function expects the style text as an argument. It's not really possible to think of all the use cases and how it'll be affecting them.

This also changes the meaning of how the returned string is treated within `css` and outside `css`. If I interpolate this class name in a regular string, I wouldn't get the CSS text, I'd get the classname, regardless of whether it's followed by a dot or not. Looking at the API, I wouldn't think that it would work like that. It can be explained in the docs, sure, but a better API should be self-explanatory.

> The only part I don't like about introduced API is the confusion behind using props interpolation in css, but I believe this can be well explained in the docs.

People don't really read docs, they skim through them if they actually try to read them. That's why invalid things should be impossible to do with the API.

Why not add a new API instead of overloading the meaning of the `css` tag? It'd make it explicit and reduce confusion. I get that making Linaria's API compatible with Emotion makes migration easier, but it's already incompatible with the way emotion's class names are used. A codemod is probably a better way to migrate.

jayu:
@satya164 so what API you propose then? 

`fragment.css` seems to not support syntax highlighting.

jayu:
@satya164 hmm ?

satya164:
idk, i'm just saying that overloading `css` isn't great. the vscode extension is for styled components, so it's probably not going to support syntax highlighting linaria specific APIs

Jayphen:
> A codemod is probably a better way to migrate.

How could a codemod reconcile the incompatible APIs? It could remove the ‘css’ call from the style declarations, leaving them as plain string template literals, but then they cannot be used alone. Afaict there’s no possible way to codemod unless a compose API is introduced

satya164:
> Afaict there’s no possible way to codemod unless a compose API is introduced

If the codemod doesn't try to look at how something is used, then sure. It could look at how a the thing is used and rewrite it accordngly.

shiro:
I also agree that `css` should not be overloaded since extensions depend on it.

I don't think there's a way to use the existing styled components API to implement fragments so extensions would support it, however some extensions (like the one for IntelliJ) provide the option to define custom tags, meaning you can get highlighting for `fragment`.

IMO either adding a new tag or using a compose function like `compose(someFragmentCSS)` (not sure if possible to implement) would be a good way to do things.

I think fragments are a very useful concept and should be added, even if not all extensions support it at first.  
Sidenote: I'm not sure about the implications on stylelint.

jayu:
That was fun to prototype!

PR #620, opened 2020-05-25
Babel plugin typings

### LINKED ISSUE #616: Optionally remove babel plugin typings from final build
## Describe the enhancement
There was an issue #599 with referencing babel plugin types from our `linaria` package.

I'd like to remove all @babel/core types dependencies from `linaria` and `linaria/react` to prevent similar issues in the future.
## Motivation
It assures that there won't be any cross references between babel plugin and linaria package.

## Possible implementations
Remove all references to @babel/core types from `linaria` and `linaria/react`.
I'd add seperate tsconfig which will exclude babel types. I'd add @types/babel__core as dev dependency because then we will be immune to changing jest for something else.

## Related Issues

#599

TMaszko:
## Motivation
Remove dependency on babel types from `core` and `react` and types declarations other than necessary for these.

## Summary
Fixes #616

## Test plan
green CI


tsconfig.json
```diff
@@ -1,13 +1,10 @@
...
-    "src/**/*.ts",
-    "src/**/*.tsx"
-  ],
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
   "exclude": [
     "node_modules",
     "__ts-tests__",
     "**/__tests__/**",
-    "**/__fixtures__/**",
+    "**/__fixtures__/**"
```
jayu, line 7: We should fix prettier config in linaria :P

PR #626, opened 2020-05-28
Transformer returns always transformed code if file contains only unused linaria code

### LINKED ISSUE #624: Transforming file contains only unused linaria code always results in original code
## Environment
- Linaria version: 2.0.0-alpha.5

## Description
File with unused code like:
```js
import { css } from "linaria";

const bar = css`
  background-color: red;
`;
```
Should be transformed to sth like:
```js
import { css } from "linaria";

const bar = 'className'
```
Then bundler should remove it later anyway using tree shaking.
If code wasn't transformed then bundler will include original implementation with error in runtime because css were treated as a function call( which can include side effects).

## Possible solution
Always return transformed code from `transformer` instead of original one

TMaszko:
## Motivation
There was a bug with transforming file that contains only unused linaria code( like variables that never were used again)

## Summary
Fixes #624
Transformer returns

## Test plan
Added tests for this case.

### DIFF
--- a/src/__tests__/transform.test.ts
+++ b/src/__tests__/transform.test.ts
@@ -212,3 +212,24 @@ it("doesn't throw due to duplicate preset", async () => {
     )
   ).not.toThrow('Duplicate plugin/preset detected');
 });
+
+it('testowy', async () => {
+  const { code } = await transform(
+    dedent`
+    import { css } from 'linaria';
+
+    const title = css\`
+      color: red;
+    \`;
+    `,
+    {
+      filename: './test.js',
+      outputFilename,
+      pluginOptions: {
+        rules,
+      },
+    }
+  );
+
+  expect(code).not.toContain('css`');
+});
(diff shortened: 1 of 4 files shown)


src/__tests__/transform.test.ts
```diff
@@ -212,3 +212,24 @@ it("doesn't throw due to duplicate preset", async () => {
     )
   ).not.toThrow('Duplicate plugin/preset detected');
 });
+
+it('testowy', async () => {
```
jayu, line 216: Please adjust the name of the test

jayu:
Regarding the TS error, I had the same once when I run typecheck locally, but it didn't happen on CI, so I ignored it. Do you have this error locally also?

TMaszko:
Yeah I had it

src/__tests__/transform.test.ts
TMaszko: 😂

PR #627, opened 2020-05-29
Removed rest operator from styled

### LINKED ISSUE #618: "Unexpected token ..." error
I occasionally see this error in my production logs:
```
Unexpected token ...
```

Sentry says it is coming from here:
```
 const { as: component = tag, class: className, ...rest } = props;
```
https://github.com/callstack/linaria/blob/eda0efaa437b267f536fa4866efd9299f9738767/src/react/styled.ts#L69

- Linaria version: 2.0.0-alpha.3
- Bundler (+ version): Next.js 9.4.0
- Node.js version: 13.11.0
- OS:  Chrome 79.0.3945 on windows 10
- User Agent:  Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.136 Safari/537.36
--

## Description

I thought this was an older browser issue at first so I ignored it but it says Chrome 79 which I think should support rest spread.

image

## Reproducible Demo

Unfortunately I can't reproduce, I just see it in Sentry error monitoring from time to time.  Is there some rest spread babel during build not happening?  I only see a few instances so I assume it must be something edge case so feel free to close.  I see there are new versions out I will update to.  Unfortunately it happens infrequently enough it will be hard to know for a while if it is fixed.  I just wanted to call it out just incase.

TMaszko:
## Motivation
There were last usage of rest operator which caused issues without additional transpliling linaria on the user side.

## Summary
Fixes #618 

## Test plan
green CI

### DIFF
--- a/src/react/styled.ts
+++ b/src/react/styled.ts
@@ -26,6 +26,15 @@ type Options = {
   };
 };
 
+// Workaround for rest operator
+const restOp = (
+  obj: { [key: string]: any },
+  keysToExclude: string[]
+): { [key: string]: any } =>
+  Object.keys(obj)
+    .filter(prop => !keysToExclude.includes(prop))
+    .reduce((acc, curr) => Object.assign(acc, { [curr]: obj[curr] }), {}); // rest operator workaround
+
 const warnIfInvalid = (value: any, componentName: string) => {
   if (process.env.NODE_ENV !== 'production') {
     if (
@@ -72,7 +81,8 @@ function styled(tag: any): any {
     }
 
     const render = (props: any, ref: any) => {
-      const { as: component = tag, class: className, ...rest } = props;
+      const { as: component = tag, class: className } = props;
+      const rest = restOp(props, ['as', 'class']);
       let filteredProps;
 
       // Check if it's an HTML tag and not a custom element
--- a/src/react/styled.ts
+++ b/src/react/styled.ts
@@ -140,7 +140,10 @@ function styled(tag: any): any {
       ? React.forwardRef(render)
       : // React.forwardRef won't available on older React versions and in Preact
         // Fallback to a innerRef prop in that case
-        ({ innerRef, ...rest }: any) => render(rest, innerRef);
+        (props: any) => {
+          const rest = restOp(props, ['innerRef']);
+          return render(rest, props.innerRef);
+        };
 
     (Result as any).displayName = options.name;
 


jayu:
I have an idea here, let's check which browser version is incompatible with rest operator, and adjust our `preset-env` settings. The test for `core-js` should fail since rest operator should be polyfilled with `core-js` 
So maybe we should change `last 2 versions` to eg. `last 10 versions`
https://github.com/callstack/linaria/blob/a3a989d3b8664da54fb5b0cda7bc59e6f7445b81/babel.config.js#L16-L23

TMaszko:
@jayu Hmm but seems that in this case user using MS Edge in the least version that is not running on chromium :/ also `last 10 versions` wouldn't be enough for the majority of browsers see https://caniuse.com/#feat=mdn-javascript_operators_spread_spread_in_destructuring

jayu:
got it

satya164:
Is this a bug in babel preset-env? Seems like our browserslist query includes that version of Edge, so it should be compiled, right?

I was checking the browserslist config and this would be the query to support all browsers that support CSS variables:

```json
'last 2 versions',
'not ie 11',
'not ie_mob 11',
'not and_qq 10',
'not baidu 7',
'not op_mini all',
'edge >= 15',
'firefox >= 31',
'chrome >= 49',
'safari >= 9.3',
'opera >= 36',
'android >= 81',
'samsung >= 5',
'not dead'
```

le0pard:
Thanks for fix. It fix linaria usage in latest Edge. But not all `..rest` part fixed. You also need to fix this part:

https://github.com/callstack/linaria/blob/master/src/react/styled.ts#L131-L133

I did quick fix directly in package and check that this finally works:

```
    ((otherProps) => {
      const {innerRef} = otherProps;
      const rest = restOp(otherProps, ['innerRef']);
      return render(rest, innerRef);
    });
```

image

What do you think @TMaszko ?

jayu:
Thanks for the note @le0pard . I will update this PR in a moment.
I checked browser list posted by Satya, we would have to get rid of following functions to support all of those browsers:
```
es.array.filter { "chrome":"49", "firefox":"31", "opera":"36" }
es.array.includes { "chrome":"49", "firefox":"31", "opera":"36", "samsung":"5" }
es.array.index-of { "chrome":"49", "firefox":"31", "opera":"36", "safari":"10" }
es.array.reduce { "firefox":"31" }
es.object.assign { "edge":"15", "firefox":"31" }
es.object.keys { "firefox":"31" }
es.string.includes { "chrome":"49", "edge":"15", "firefox":"31", "opera":"36" }
```
But I think it is not worth to mess the file more until someone reports the crash.

SpawnAtis:
Any updates ?

wereHamster:
@SpawnAtis I suggest that you configure your project to process linaria (or all dependencies, if that's easier to configure) through babel (or another transpiler that your project uses).

PR #631, opened 2020-06-19
docs: update preact configuration

tamorim:
## Motivation

Easier setup of `linaria` with `preact` when coming from a `preact-cli` generated boilerplate.

#590 

## Test plan

Generate a new project using the `preact-cli` and follow the updated steps and ensure that's working as intended.


jayu:
Thanks for the PR @tamorim ! 
Can you please also upgrade the config in out preact example https://github.com/callstack/linaria/tree/master/examples/Preact/preact-example ?

Thanks to that we will be able to verify if the change works!

tamorim:
@jayu Sorry for missing the example 😅 
Just pushed the new changes, thanks for the direction!

PR #632, opened 2020-06-22
Mark package as side-effect free

wereHamster:
## Motivation

If a package is marked side-effect free, bundlers are free to drop import declarations which import unused bindings.

For example, without this option enabled, the import declaration `import { css } from linaria` may be retained in the bundler even if all calls to `css` are removed by the babel plugin.

## Test plan

Check that the `css` function is removed from the bundles. We'd need rollup or webpack (or another bundler which supports the `sideEffects` option), but I'm unsure if the current test setup is prepared for that.

Update: I didn't find a nice way to write a test for that without adding a lot of complexity to the project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "linaria",
   "version": "2.0.0-alpha.5",
   "description": "Blazing fast zero-runtime CSS in JS library",
+  "sideEffects": false,
   "main": "lib/index.js",
   "module": "esm/index.js",
   "bin": "bin/linaria.js",


jayu:
Hey, thank you for that idea. Since this is a thing that has to be discussed and tested before we merge, in that kind of cases please open an issue with enhancement proposal first (but I'm good with just a PR, in this case, to not introduce redundant "paper work". We have to validate it somehow, so any demo is more than welcomed. Maybe someone could spin up a repo that uses linaria and webpack with proper configuration, and then we can modify Linaria package.json with https://www.npmjs.com/package/patch-package. Once we confirm it works as expected, a few sentences of documentation should be added to this PR and any test if possible (I'm good with not adding test if it brings too much complexity)

Issue #622, opened 2020-05-26
Typings not working when overriding another `styled` component


OrcaXS:
## Environment

Linaria version: 2.0.0-alpha.5
Bundler (+ version): webpack 4.43.0
Node.js version: v12.16.3
OS: macOS 10.15.4
React: 16.13.1

## Description
Annotating props of a `styled` component works fine like in #536:
```tsx
const Wrapper = styled.div<{ compose: boolean }>`
  width: 1em;
  background-color: ${(props) => (props.compose ? 'transparent' : 'green')};
`
```
But doesn't work on an overrode one:
```tsx
const NewWrapper = styled(Wrapper)<{ compose: boolean }>`
  width: 2em;
  background-color: ${(props) => (props.compose ? 'transparent' : 'green')};
`
```

Errors are like below:
```
Argument of type '(props: any) => any' is not assignable to parameter of type 'StaticPlaceholder'.
  Property '__linaria' is missing in type '(props: any) => any' but required in type 'StyledMeta'.ts(2345)
types.ts(47, 3): '__linaria' is declared here.
```

Am I annotating the types the wrong way? Thanks.

## Reproducible Demo
```tsx
import React from 'react';
import { styled } from 'linaria/react';

const Wrapper = styled.div<{ compose: boolean }>`
  width: 1em;
  background-color: ${(props) => (props.compose ? 'transparent' : 'green')};
`

const NewWrapper = styled(Wrapper)<{ compose: boolean }>`
  width: 2em;
  background-color: ${(props) => (props.compose ? 'transparent' : 'red')};
`

const App = () => (
  <NewWrapper compose>Test</NewWrapper>
);
```

TMaszko:
Hi! Could you provide a complete repo where we can reproduce this bug :)?`
Thanks in advance !

OrcaXS:
Sure, here it is! 
https://github.com/OrcaXS/linaria-issue-reproduce/tree/typescript-styled-override


Running `yarn tsc` throws:
```
src/index.tsx:11:23 - error TS2345: Argument of type '(props: any) => "transparent" | "red"' is not assignable to parameter of type 'StaticPlaceholder'.
  Property '__linaria' is missing in type '(props: any) => "transparent" | "red"' but required in type 'StyledMeta'.

11   background-color: ${(props) => (props.compose ? 'transparent' : 'red')};
[...]

Issue #536, opened 2020-02-04
Typings not working properly


zaguiini:
## Environment

Create react app (ejected) + TypeScript, nothing special there.

## Description

I'm trying to annotate props like this:

```ts
import { styled } from 'linaria/lib/react'

interface TitleProps {
  background: string
}

export const Title = styled.div<TitleProps>`
  background: ${(props) => props.background};
`
```

But I get an error:

```
interface TitleProps
    Expected 0 type arguments, but got 1.ts(2558)
 ```

Which is weird because the typings on the [repository](https://github.com/callstack/linaria/blob/master/react.d.ts#L13) says one thing, but the typings from the distribution package at `linaria/lib/react/styled.d.ts@17` says:

```ts
declare type HtmlStyledTag<TName extends keyof JSX.IntrinsicElements> = (strings: TemplateStringsArray, ...exprs: Array<StaticPlaceholder | ((props: JSX.IntrinsicElements[TName]) => string | number)>) => StyledComponent<JSX.IntrinsicElements[TName]>;
```

Which means that it's not accepting any generics on this besides `TName`. It works if I do this:

```ts
declare type HtmlStyledTag<TName extends keyof JSX.IntrinsicElements> = <AdditionalProps>(strings: TemplateStringsArray, ...exprs: Array<StaticPlaceholder | ((props: JSX.IntrinsicElements[TName] & AdditionalProps) => string | number)>) => StyledComponent<JSX.IntrinsicElements[TName] & AdditionalProps>;
```

(`AdditionalProps` added)

But it looks like this file is being auto-generated and it looks that importing from `linaria/lib/react` is a workaround, right? What can be done to address this issue?

Anber:
Looks like it's time to merge #389 :)

Anber:
@zaguiini thank you for your report! The fix will be released with the next beta.

PR #639, opened 2020-07-11
fix: typings for additional props (#622)

Anber:
## Motivation

This PR fixes issues #622 and #536

## Test plan

The new case was added to dtslint tests and `test:dts` was added to CI for avoiding future regressions.

### DIFF
--- a/types/styled.ts
+++ b/types/styled.ts
@@ -84,16 +84,38 @@ styled.a`
   }
 `({ href: 'about:blank' });
 
-// Issue #536
-const Title = styled.div<{ background: string }>`
-  background: ${props => props.background};
-`;
+((/* Issue #536 */) => {
+  const Title = styled.div<{ background: string }>`
+    background: ${props => props.background};
+  `;
 
-// $ExpectType "extends"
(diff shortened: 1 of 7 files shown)


types/styled.ts
```diff
@@ -84,16 +84,38 @@ styled.a`
...
+
+  const styledTag = styled(Wrapper);
+
+  const NewWrapper = styledTag<{ prop2: string }>`
+    width: 2em;
+    background-color: ${props => (props.prop1 ? 'transparent' : 'red')};
+    color: ${props => props.prop2};
+  `;
+
+  // $ExpectType Validator<boolean> | undefined
```
jayu, line 116: Shouldn't it be always a `boolean` if prop1 is required? However, I'm new to dtslint so I may misunderstand syntax :P

Anber: It's not directly about a field type, it's about a `propTypes` which can either be a `Validator<something>` or `undefined`. If `prop1` is `boolean | undefined`, `propTypes` should be `Validator<boolean | undefined> | undefined`. It's not really clear, but afaik it's the simplest way to check the property type :)

PR #640, opened 2020-07-18
fix: push '\' into className before regex special characters

SeokminHong:
## Motivation

When I use a HTML elements containing regex's special characters(`\`, `$`, and so on) for class, `extractClassesFromHtml` function fails to parse them.

## Summary

`extractClassesFromHtml` returns a regex array for matching class, thus I added escape characters(`\`) before the special characters.

## Test plan

Try to parse the HTML document having class containing following characters for its class:

`\`, `^`, `$`, `{`, `}`, `[`, `]`, `(`, `)`, `.`, `*`, `+`, `?`, `|`

For example: `<code class="language-C++">`


iamakulov:
Just stumbled upon the same error! Would love to see this merged. How can I help?

Anber (approved):

PR #644, opened 2020-07-29
feat: tightened typing for Object Interpolation (CSSProperties)

shqld:
## Motivation

[Object Interpolation](https://github.com/callstack/linaria/blob/master/docs/BASICS.md#object-interpolations) is very useful especially for creating _sass-like mixins_ such as

```ts
function margin(px: number): CSSProperties {
  return { margin: px + 'px' };
}

const divStyle = css`
  ${margin(12)}
`;
```

However, currently `CSSProperties` type accepts any keys and values even outside of CSS rule stuff. This is problematic for type safety and usability (e.g. auto-completion).

## Summary

In this PR, I've just replaced declaration of `CSSProperties`.

Concretely, replaced `[key: string]: string | number` with `csstype.Properties` which is responsible for covering all available key-value pairs of css property.

This allows us more relialbe type cheking and usable auto-completion for Object Interpolation without downside in runtime.

For extendability of property types, internal interface called `Properties` is exported so that users can merge interface at their discretion in the user-land.

## Test plan

```ts
function marginTop(px: number): CSSPropreties {
  return {
    // OK
    marginTop: px + 'px',
    // OK
    'margin-top': 0,
    // ERROR
    'margin-top': 1,
    // ERROR
    'my-prop': 'hello',
  };
}

css`
  ${{
    // OK
    marginTop: px + 'px',
    // OK
    'margin-top': 0,
    // ERROR
    'margin-top': 1,
    // ERROR
    'my-prop': 'hello',
  }}
`;
```

```ts
declare module 'linaria/lib/core/css' {
  interface Properties {
    'my-prop': 'hello';
  }
}

function marginTop(px: number): CSSPropreties {
  return {
    // OK
    'my-prop': 'hello',
  };
}

css`
  ${{
    // OK
    'my-prop': 'hello',
  }}
`;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
     "@emotion/is-prop-valid": "^0.7.3",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "cosmiconfig": "^5.1.0",
+    "csstype": "^2.6.11",
     "debug": "^4.1.1",
     "dedent": "^0.7.0",
     "enhanced-resolve": "^4.1.0",
--- a/src/core/css.ts
+++ b/src/core/css.ts
@@ -1,8 +1,18 @@
+// 'csstype' cannot be resolved since 'main' field in its package.json is empty and no script files are exported
+// eslint-disable-next-line import/no-unresolved
+import * as csstype from 'csstype';
 import { StyledMeta } from '../StyledMeta';
 
-type CSSProperties = {
-  [key: string]: string | number | CSSProperties;
-};
+// exported so that users can extend the interface by declaration merging
+export interface Properties
+  extends csstype.Properties,
+    csstype.PropertiesHyphen {}
+
+export type CSSProperties =
+  | Properties
+  | {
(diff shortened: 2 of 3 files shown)


Anber:
Hi @shqld!
Unfortunately, `csstype` doesn't work with CSS custom properties and it can be a problem for us. The problem looks unresolvable until [that proposal](https://github.com/microsoft/TypeScript/issues/6579) isn't implemented.

shqld:
Hi @Anber ！Thanks for your reply.

We can specify any types we want by overriding types using TypeScript's Declaration Merging (as I described above), so I believe it will still have some effect to do something like in this PR without special support for custom properties. At least no disadvantage is considered, IMHO.

Anber:
Hi @shqld!
There is good news. TypeScript 4.1 introduces template literal types so that it will probably become possible to write the new CSSProperties type which can handle custom properties. Let's wait for the 4.1 release.

Issue #643, opened 2020-07-27
Webpack and multiple babel modules are production dependencies


tommoor:
## Environment

- Linaria version: 2.0.0-rc.0"
- Bundler (+ version): Yarn 1.22.4
- Node.js version: v12.16.1
- OS: macOS

## Description

Webpack and multiple babel modules are marked as production dependencies, inflating build sizes and production install times.

See package.json: https://github.com/callstack/linaria/blob/master/package.json#L125

```
=> Found "watchpack#fsevents@2.1.3"
info Reasons this module exists
   - "remirror#@remirror#core#linaria#webpack#watchpack#chokidar" depends on it
   - Hoisted from "@elasticprojects#abstract-core#remirror#@remirror#core#linaria#webpack#watchpack#chokidar#fsevents"
info Disk size without dependencies: "56KB"
info Disk size with unique dependencies: "56KB"
info Disk size with transitive dependencies: "56KB"
info Number of shared dependencies: 0
```

ifiokjr:
Just to give some context, I'm the maintainer of [`remirror`](https://github.com/remirror/remirror) and in a recent update I've switched from using `emotion` to `linaria`. 

For the most part this has been great, but I need to also include linaria as a runtime dependency which means consumers of my library are also installing it and all it's dependencies like `webpack`, `babel`, `acorn`. 

This has caused a few issues with dependency clashes etc. For example, the latest version of `nextjs` is using webpack@5 and can't coexist with `linaria`. 

As a library author my choices are:

- drop the linaria dependency from `remirror` and roll an alternative (I'd really rather not since this library is so good!)
- ask that you split the functionality into separate packages for the build tools and the runtime code. `@linaria/core`, `@linaria/react` for runtime and `linaria` for the build tools.
- ask that you convert the build tool dependencies to `peerDepedencies`.
- suck it up (who needs happy users anyway).

I'm not sure how to proceed, but maybe others consuming this tool as part of a library are facing similar issues.

satya164:
webpack should be a peer dep. can you send a PR for that?

regarding deps like babel, the only way is to split up the package to multiple parts such as webpack loader, babel plugin and the core API. which we plan to do in future.

wereHamster:
[...]

PR #654, opened 2020-08-11
fix(deps): drop a few unnecessary dependencies (#643)

Anber:
## Motivation

We have some dependencies which we don't really use or can avoid using: webpack, @babel/core, etc.
Another problem here, that Linaria babel plugin uses an external version of babel instead of a version which launches the plugin, that can probably cause some tricky errors.

## Summary

Despite the fact that PR touches a lot of files, the most changes are just `type` annotations for imports and linted files (update of eslint was required for supporting `type`). The real change here is passing `babel` through call stacks in the plugin, but it doesn't affect the logic.
There are also three minor fixes:
* src/babel/extract.ts — better typings for `expressionsToEvaluate` and `originalLazyExpressions`;
* src/babel/visitors/CollectDependencies.ts — fix for a bug with hoisting in different versions of babel;
* src/babel/visitors/GenerateClassNames.ts — small typings improve.

Results:
* webpack has gone;
* @babel/traverse and @babel/register have gone;
* @babel/core has been moved to peerDependecies.

## Test plan

Since the logic wasn't affected, all existed tests are still working.


src/__tests__/evaluators/shaker.test.ts
```diff
@@ -1,22 +1,21 @@
...
-  only: string[] = ['__linariaPreval']
-) {
+function _shake(opts?: TransformOptions, only: string[] = ['__linariaPreval']) {
   return (
     literal: TemplateStringsArray,
     ...placeholders: string[]
   ): [string, Map<string, string[]>] => {
     const code = dedent(literal, ...placeholders);
     const [shaken, deps] = shake(
+      babel,
```
thymikee, line 18: what's the benefit of injecting Babel here? I don't think we'll inject various versions of Babel and it stays as dependency, but peer. I think we should be fine by leaving these functions as they were – dependent on `@babel/core`. This would shrink the diff quite a bit :)

package.json
```diff
@@ -53,16 +53,15 @@
...
-    "@babel/preset-react": "^7.9.4",
-    "@babel/preset-typescript": "^7.9.0",
-    "@callstack/eslint-config": "^9.1.0",
+    "@babel/cli": ">=7",
+    "@babel/plugin-proposal-class-properties": ">=7",
+    "@babel/plugin-syntax-jsx": ">=7",
+    "@babel/preset-env": ">=7",
+    "@babel/preset-react": ">=7",
+    "@babel/preset-typescript": ">=7",
+    "@callstack/eslint-config": "^10.0.1",
```
thymikee, line 62: can we upgrade eslint config and prettier in a separate PR? they add a lot of unnecessary noise

docs/HOW_IT_WORKS.md
```diff
@@ -174,6 +174,7 @@ Sometimes it can be useful to implement your own strategy (it can be just a mock
 
 ```typescript
 type Evaluator = (
+  babel: BabelCore,
```
thymikee, line 177: no need for this change imho, as stated somewhere earlier

package.json
Anber: Without the upgrade, it's impossible to use `import type`, and without `import type` it's impossible to use peerDependencies :)

However, I can move all these eslint things to another PR which will be a pre-requirement for that PR.

thymikee: Yes please :)

src/__tests__/evaluators/shaker.test.ts
Anber: Hm… We don't touch the original AST in evaluators, so it's probably safe to use imported babel here. I'll take a close look.

wereHamster:
Why do we need type-only imports?

https://twitter.com/drosenwasser/status/1218253612746301440

Anber:
@wereHamster mostly because of linter: eslint doesn't complain about type-only imports if dependencies aren't specified in package.json. Another reason is improved readability, that's help to manage dependencies.

docs/HOW_IT_WORKS.md
thymikee: Looks like this is the last thing to revert now :)

Issue #657, opened 2020-08-15
Linaria website is not available on now deployment


marvinhagemeister:
I just get a blank page with this message instead of the linaria website:

![Screenshot 2020-08-15 at 10 10 51](https://user-images.githubusercontent.com/1062408/90308418-a5dcb680-dedf-11ea-86dd-55fb01cf0e3b.png)

thymikee:
@satya164 mind having a look?

jayu:
Since we have custom node.js server there, we have to migrate to some static solution. Next.js static generated pages could be an option for us.

PR #660, opened 2020-08-19
chore: basic next.js setup

jayu:
## Motivation

https://github.com/callstack/linaria/issues/657
Vercel no longer support the deployment of pages with runtime, so we decided to transform landing to static page built with next.js

## Summary

Todo List

- [x] basic next.js setup
- [x] create `_app.js` and import global CSS from `website/index.js` 
- [x] create `_document.js` and add proper HTML head properties
- [x] verify if SSG works
- [x] remove redundant deps, files, scripts
- [x] deploy landing 

Note: Currently landing pages use local version of linaria since it might be a possible blocker for a proper build, so we may consider using linaria from npm


grabbou:
I have managed to deploy this PR under https://linaria-new.grabbou.vercel.app/ with some changes (see my commit).

Question to @simka and @jayu - what do we need `webpack.config.js` for? Isn't that overlapping with `next.config.js`?

website/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "@zeit/next-css": "^1.0.1",
+    "linaria": "^1.3.3",
```
grabbou, line 15: Ideally, if the project was `Yarn workspaces`, one could use a local version while debugging, but I don't think this will be a major issue for now (?)

grabbou:
I will set up CI in a while (to automatically deploy) and link it with current `linaria.now.sh`. Will let you know when ready 👊

PR #669, opened 2020-09-01
fix(typings): styled mistakenly infers type from interpolated component

Anber:
## Motivation

Fix for #622 introduces regression in `styled`:
```
const Header = styled.h1``;
const Block = styled(/* some custom component  */)`
  ${Header} {
    /* styled */
  }
`; // type of Block here is inferred from Header and not from the original component
```

## Test plan

The new case for dtslint was added.


jayu (approved):
I trust it works :D

PR #670, opened 2020-09-03
fix(evaluators): respect environment variables during evaluation

### LINKED ISSUE #666: Environment variables not respected during Interpolation
## Environment

- Linaria version: 2.0.0-alpha.5
- Bundler (+ version): 4.42.1
- Node.js version: 12.14.1
- OS: macOS 10.15.6

## Description

I have a project that uses Linaria and I want to set the theme at build time using an environment variable. The Theme.js file goes something like this:

```js
const theme = process.env.REACT_APP_THEME_NAME === 'MAIN'
    ? require('../theme-variables.js')
    : require('../theme-variables-alt.js')

export { theme };
```

and there is a Button.js file that goes like this

```js
import { styled } from 'linaria/react';
import { theme } from '../Theme';
const Button = styled('button')`
  background-color: ${theme.colors.blue};
`;
```

However, for some reason during interpolation, the complied CSS will only contain the values from the alt file despite the `REACT_APP_THEME_NAME` environment variable being set to `MAIN`. 

## Reproducible Demo

https://codesandbox.io/s/linaria-demo-forked-q25fw?file=/src/App.js

Anber:
## Motivation

It fixes #666

## Summary

It looks logical to pass all env variables and not only NODE_ENV. I don't know why we had such a limitation.

### DIFF
--- a/src/babel/process.ts
+++ b/src/babel/process.ts
@@ -28,6 +28,4 @@ export const memoryUsage = noop;
 export const uvCounters = noop;
 export const features = {};
 
-export const env = {
-  NODE_ENV: process.env.NODE_ENV,
-};
+export const env = process.env;


woahitsraj:
I actually tried this approach when I attempted to fix it myself and it didn't seem to work but I may have gotten something wrong. Before this gets merged in maybe it could be verified with this minimal failing example I set up?

https://codesandbox.io/s/linaria-demo-forked-q25fw?file=/src/App.js

jayu (approved):
lgtm

Issue #413, opened 2019-05-16
Migrate Linaria to monorepo under @linaria namespace


thymikee:
## Describe the feature

Let's split Linaria to monorepo after #386 is done.
We already grabbed the namespace: http://npmjs.com/org/linaria

## Motivation

Splitting to smaller packages, like `@linaria/core` , `@linaria/runtime`, `@linaria/react` etc is going to allow adding e.g. library's runtime as a direct dependency of other libraries that take size into consideration.

PR #687, opened 2020-10-28
refactor: split one `linaria` package into multiple `@linaria/*`

Anber:
## Motivation

Closes https://github.com/callstack/linaria/issues/413

## Summary

The list of packages:

- @linaria/babel
- @linaria/cli
- @linaria/core
- @linaria/extractor
- @linaria/logger
- @linaria/preeval
- @linaria/react
- @linaria/rollup
- @linaria/server
- @linaria/shaker
- @linaria/stylelint
- @linaria/webpack4-loader


docs/CONFIGURATION.md
```diff
@@ -129,7 +129,10 @@ After that, your `package.json` should look like the following:
     "sirv-cli": "^0.4.5"
   },
   "dependencies": {
-+   "linaria": "^1.3.3",
```
satya164, line 132: Is this needed?

docs/CONFIGURATION.md
```diff
@@ -94,7 +94,7 @@ To use this preset, add `linaria/babel` to your Babel configuration at the end o
   "presets": [
     "@babel/preset-env",
     "@babel/preset-react",
-+   "linaria/babel"
```
satya164, line 97: Seems leftover

package.json
```diff
@@ -1,12 +1,13 @@
 {
+  "private": true,
   "name": "linaria",
   "version": "2.0.0",
   "description": "Blazing fast zero-runtime CSS in JS library",
   "main": "lib/index.js",
```
satya164, line 6: I think we can remove all publishing related fields from the root package.json

package.json
```diff
@@ -66,80 +65,41 @@
     "@types/debug": "^4.1.5",
     "@types/dedent": "^0.7.0",
     "@types/enhanced-resolve": "^3.0.6",
```
satya164, line 67: can we move type dependencies to the relevant packages which need them? e.g. to the webpack loader for enahanced resolve

packages/babel/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["src/**/*.ts", "src/**/*.tsx"]
```
satya164, line 3: Is this `include` needed? Shouldn't it include all files by default?

We also might want to use references, e.g.:

```json
{
  "extends": "../../tsconfig",
  "references": [
    { "path": "../core" }
  ],
  "compilerOptions": {
    "outDir": "./types"
  }
}
```

satya164 (commented):
We also might want to configure `paths` in our `tsconfig.json` so that typescript uses the source files when typechecking:

```js
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@linaria/*": ["./packages/*/src"]
    },
    "composite": true,
    "allowUnreachableCode": false,
    "allowUnusedLabels": false,
    "esModuleInterop": true,
    "importsNotUsedAsValues": "error",
    "forceConsistentCasingInFileNames": true,
    "jsx": "react",
    "lib": ["esnext", "dom"],
    "module": "esnext",
    "moduleResolution": "node",
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "noImplicitUseStrict": false,
    "noStrictGenericChecks": false,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "resolveJsonModule": true,
    "strict": true,
    "target": "esnext"
  }
}
```

Though I think we'd need separate build configs for TS in individual packages if we do it:

```json
{
  "extends": "./tsconfig",
  "compilerOptions": {
    "paths": {}
  }
}
```

docs/CONFIGURATION.md
Anber: What is needed? `"linaria": "^1.3.3"` here was replaced with a bunch of `@linaria` packages.

docs/CONFIGURATION.md
Anber: You have probably been confused by this plus sign at the beginning of the line, but this line was removed :)

package.json
Anber: Lerna suggests keeping dev dependencies on the top level, but it makes sense only if a dependency is used in 2+ packages. However, it has automatically collected deps from all packages :)

Anber:
Hi @satya164!
I hope I got it right :)

packages/cli/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "paths": {}
```
thymikee, line 4: If a package references other internal package, it's good to set project references as well to provide deterministic builds: https://www.typescriptlang.org/docs/handbook/project-references.html. We use this at Jest monorepo

Anber: Is it really needed if we don't use tsc for building? Lerna itself checks the usages and provides deterministic builds.

package.json
Anber: Oops. Forgot to clear that mess.

package.json
```diff
@@ -1,57 +1,25 @@
...
-    "typecheck": "tsc --noEmit",
-    "test": "jest",
-    "test:dts": "dtslint --localTs node_modules/typescript/lib types",
+    "check:all": "yarn test:dts && yarn typecheck && yarn test && yarn lint",
     "add-contributor": "all-contributors add",
-    "build:lib": "NODE_ENV=legacy babel src --out-dir lib --extensions '.js,.jsx,.ts,.tsx' --ignore '**/__tests__/**,**/__integration-tests__/**,**/__fixtures__/**' --source-maps --delete-dir-on-start",
-    "build:esm": "babel src --out-dir esm --extensions '.js,.jsx,.ts,.tsx' --ignore '**/__tests__/**,**/__integration-tests__/**,**/__fixtures__/**' --source-maps --delete-dir-on-start",
-    "build": "yarn build:lib && yarn build:esm",
-    "build:declarations": "tsc --project tsconfig.prod.json --emitDeclarationOnly --outDir lib",
     "watch": "yarn build --watch",
```
thymikee, line 18: there's no `build` script anymore

package.json
```diff
@@ -1,57 +1,25 @@
...
-    "css-in-js",
-    "styled-components",
-    "babel-plugin",
-    "babel",
-    "webpack"
-  ],
   "scripts": {
+    "prepare": "yarn lerna run prepare",
+    "test": "yarn lerna run test",
+    "test:coverage": "yarn lerna run test -- -- --coverage",
```
thymikee, line 12: Do we really need these extra `--`?
```suggestion
    "test:coverage": "yarn lerna run test --coverage",
```

packages/cli/tsconfig.json
satya164: > if we don't use tsc for building
we'd still need to use tsc for building type definitions

docs/CONFIGURATION.md
satya164: yes, my bad

jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  testEnvironment: 'node',
+  testRegex: '/__tests__/.*\\.(test|spec)\\.(js|tsx?)$',
+  transform: {
+    '^.+\\.(js|ts|tsx)$': 'babel-jest',
```
thymikee, line 5: we can probably remove the `transform` and `testRegex` since defaults cover what we use

package.json
satya164: I guess by dev deps it means tooling like esint prettier etc. otherwise it's strange to have a dependency in one place and the type that complements it in a another place.

packages/cli/tsconfig.json
Anber: Yeah, but it's also called by Lerna in a predefined order.

package.json
Anber: Unfortunately. The first one says that the rest of arguments should be ignored by yarn. Usually, it can be simply ignored, since yarn passes by all unknown arguments, but the following `--` is not an unknown.

package.json
Anber: It suggests it mostly because of, in that case, it is easier to maintain deps versions between packages.

satya164: imo it's harder to maintain when you have one dependency in a package, and it's corresponding type in another place. when you change version of the package, you also need to remember to change the type to the corresponding version in a totally different place.

Anber: I definitely agree that `@types` should be in the same package.json as the dependencies, but what do you think about `@babel`-packages? They are using in almost all packages and I think it would be better to keep them in the root package.json

packages/cli/tsconfig.json
Anber: Seems like https://www.npmjs.com/package/@azu/lerna-to-typescript-project-references can do it automatically. Maybe just add it to `prepare` step?

Anber:
I'm ready for the next wave! :)

Anber:
Oh… well. Something went wrong with references.

thymikee (approved):
🚀

antitoxic:
👋 Hey @Anber what the status of that split? I see the webpack docs https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#webpack refer to A non-existing package called `@linaria/webpack-loader`.

After some digging in npm I can find `@linaria/webpack4-loader`, but trying to set this up is blowing up with some babel duplicate plugins errors.

Anber:
Hi @antitoxic 

It's still in the beta. `@linaria/webpack-loader` is reserved for webpack 4/5 loader and will be released later this month.

> trying to set this up is blowing up with some babel duplicate plugins errors.
It should work. Could you please create an issue with details?

PR #695, opened 2020-11-08
fix: fix typo in the error message

roottool:
## Motivation

Typo in the unexpected runtime error message is fixed by this PR.

## Test plan

I tried `yarn run check:all`. The result is no error.


Anber (approved):

Anber:
Hi @roottool!

> I tried `yarn run check:all`. The result is no error.

It's strange that we don't have a test for this case.

PR #709, opened 2020-11-25
Refine preset-env targets to support more browsers

PierreGUI:
## Motivation
Looking at my logs in Sentry I found out many users are getting `SyntaxError` originating from Linaria's source code.  
Here is just one example: `SyntaxError: Unexpected token =>  at ? (./node_modules/linaria/lib/react/styled.js:26:1)`

That's because in some older browsers arrow functions are not supported: https://caniuse.com/arrow-functions  

## Summary
I assume Linaria is possibly compatible with **all browsers with CSS Variables support**: https://caniuse.com/css-variables
By taking this requirement into `targets`, we make sure Linaria builds into a JS library that works for all supported browsers.  
Babel will automatically find out all the necessary `transform` plugins, eg: `transform-arrow-functions { "firefox":"41", "ios":"9.3", "safari":"9.1", "samsung":"4" }` (from babel debug output).

## Test plan
Example of a browser that supports CSS Variables but not arrow functions: Safari 9.1 (until Safari 10)

Another example is `const declaration`:  
Error: `SyntaxError: Unexpected keyword 'const'.`  
Compatibility: https://caniuse.com/const

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -12,18 +12,19 @@
 /*
  * Configuration for the legacy build
  * Browsers compatibility based on https://caniuse.com/css-variables
+ * Versions adjusted specifically to not include any core-js polyfills
  */
 
 const commonJSTargets = {
   browsers: [
-    'Chrome >= 49',
-    'ChromeAndroid >= 49',
-    'Edge >= 16',
-    'Firefox >= 31',
-    'iOS >= 9.3',
-    'Opera >= 36',
-    'Safari >= 9.1',
-    'Samsung >= 5',
+    'Chrome > 52',
+    'ChromeAndroid > 52',
+    'Edge > 18',
+    'Firefox > 49',
+    'iOS > 10.3',
+    'Opera > 39',
+    'Safari > 10.1',
+    'Samsung > 5',
     'last 2 versions',
     'not ie 11',
     'not ie_mob 11',
--- a/babel.config.js
+++ b/babel.config.js
@@ -11,10 +11,19 @@
 
 /*
  * Configuration for the legacy build
+ * Browsers compatibility based on https://caniuse.com/css-variables
  */
 
 const commonJSTargets = {
   browsers: [
+    'Chrome >= 49',
+    'ChromeAndroid >= 49',
+    'Edge >= 16',
+    'Firefox >= 31',
+    'iOS >= 9.3',
+    'Opera >= 36',
+    'Safari >= 9.1',
+    'Samsung >= 5',
     'last 2 versions',
     'not ie 11',
     'not ie_mob 11',


Anber (approved):

Anber:
Hi @PierreGUI 

It's not so simple…
If we want to support old browsers, we have to refactor `core` and `styled` in order to exclude implicit `core-js`. Otherwise, our "zero-runtime" will be far from zero :)

PierreGUI:
Hey @Anber 
Fair point, it's possible this new set of requirements for `targets` induces the need for `core-js` polyfills. Haven't checked that to be honest (will do now).

What I had in mind are the additional `transforms` plugins from babel to transpile the code into a more broadly compatible syntax, therefore addressing these `SyntaxError`.

PierreGUI:
Ehm, sorry about that. Tests are indeed failing because Babel added this ONE polyfill 🤣 : `[./packages/core/src/cx.ts] Added following core-js polyfill: es.array.filter { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }`

On the other hand, the list of additional Babel transform plugins looks pretty nice:
```
transform-parameters { "edge":"16", "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-dotall-regex { "chrome":"49", "edge":"16", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1", "samsung":"5" }
transform-named-capturing-groups-regex { "chrome":"49", "edge":"16", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1", "samsung":"5" }
transform-async-to-generator { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1", "samsung":"5" }
transform-exponentiation-operator { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1", "samsung":"5" }
transform-template-literals { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-literals { "firefox":"31" }
transform-function-name { "chrome":"49", "edge":"16", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }
transform-arrow-functions { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-block-scoped-functions { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-classes { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-object-super { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-shorthand-properties { "firefox":"31" }
transform-duplicate-keys { "firefox":"31" }
transform-computed-properties { "firefox":"31" }
transform-for-of { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }
transform-sticky-regex { "ios":"9.3", "safari":"9.1" }
transform-unicode-escapes { "firefox":"31" }
transform-unicode-regex { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }
transform-spread { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-destructuring { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }
transform-block-scoping { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-new-target { "firefox":"31", "ios":"9.3", "safari":"9.1" }
transform-regenerator { "chrome":"49", "firefox":"31", "ios":"9.3", "opera":"36", "safari":"9.1" }
```

What if we could simply get rid of that one polyfills and get all the nice transforms, do you agree it'd be beneficial?

Anber:
Just one for `@linaria/core`, but a lot of for `@linaria/react`:
```js
require("core-js/modules/es.array.filter");
require("core-js/modules/es.array.includes");
require("core-js/modules/es.array.index-of");
require("core-js/modules/es.array.reduce");
require("core-js/modules/es.object.assign");
require("core-js/modules/es.object.keys");
require("core-js/modules/es.string.includes");
```

Actually, I don't understand why it includes all of them. `indexOf`, `filter`? All those features have been supported for ages. Look like it tries to support IE9.

So, as I mentioned before, it's not so simple :)

PierreGUI:
Yes I was wondering the same, even went to peek at Babel's source code for `present-env` without luck (the mapping for `array.filter` looks correct and targets older browsers than the one I specified 🤷 )... I'll experiment a bit and see if I can work around it.

If I can come up with a `targets` array that adds transforms but no polyfills, then we have a better idea what's the minimal browsers Linaria supports out of the box (and without refactor). It's something!

PierreGUI:
I've been digging, and here are my learnings:
- The mapping `polyfill` => `browser version` can be found here: [core-js-compat/src/data.js](https://github.com/zloirock/core-js/blob/master/packages/core-js-compat/src/data.js)
- `core-js` includes the polyfill [as long as the browser implementation does not fully comply with standard](https://github.com/zloirock/core-js/issues/644), that explains why `array.filter` is included until Chrome v51, although compatibility is [assumed from v4](https://caniuse.com/mdn-javascript_builtins_array_filter).
- A `targets` array that complies with the `no added core-js` requirement (using `data.js` mentioned above) looks like this:
```js
[
    'Chrome >= 53',
    'ChromeAndroid >= 53',
    'Firefox >= 50',
    'iOS >= 11',
    'Safari >= 11',
    'last 2 versions',
    'not ie 11',
    'not ie_mob 11',
    'not op_mini all',
    'not dead',
  ]
```

We can't say these are the minimum browsers versions Linaria is compatible with, as it probably is compatible with even earlier version, but at least it requires no extra polyfills while adding these transforms:
```
transform-parameters { "firefox":"50" }
transform-dotall-regex { "chrome":"53", "firefox":"50", "ios":"11", "safari":"11" }
transform-named-capturing-groups-regex { "chrome":"53", "firefox":"50", "ios":"11", "safari":"11" }
transform-async-to-generator { "chrome":"53", "firefox":"50" }
transform-exponentiation-operator { "firefox":"50" }
transform-template-literals { "ios":"11", "safari":"11" }
transform-literals { "firefox":"50" }
transform-function-name { "firefox":"50" }
transform-for-of { "firefox":"50" }
transform-unicode-escapes { "firefox":"50" }
transform-unicode-regex { "ios":"11", "safari":"11" }
transform-destructuring { "firefox":"50" }
transform-block-scoping { "firefox":"50" }
transform-regenerator { "firefox":"50" }
```

Although it'd already be "better", it wouldn't be enough to fix the `SyntaxError` my users are facing (missing `arrow function` transform, for example).  



#### Proposal
We could build the library with the `targets` array I suggested based on `CSS variables` compatibility.  
To avoid any polyfills from being added, set babel preset's `useBuiltIns` to `entry`.  
In `entry` mode, no `core-js` will be included but Babel will use nice transforms to fix as much `SyntaxError` as possible.  

My personal opinion is that `react/styled` contains a small runtime and therefore should be polyfilled to increase its backward-compatibility. Right now it crashes on some browsers and we don't really know what the minimum support is.

Anber:
Let's try the "no added core-js" set of rules, and then I'll rewrite `cx` and `styled` to support old browsers without core-js.

PierreGUI:
Fair enough!  
I've updated the `targets` array and added a comment too.

I rebased my branch on `2.0.x` and updated this PR

I can help with rewriting `cx` and `styled`, I guess we could simply rewrite `array.filter` etc.. as util functions?

Anber:
Thank you!

> we could simply rewrite array.filter etc.. as util functions

`cx` can be reduced to just one for-cycle. In `styled` `filter` and `reduce` can also be simplified, `includes` can be replaced with `indexOf` but replacements for the rest of methods will be tricky.
First of all, we should understand which browsers we are loosing now.

PierreGUI:
Very welcome!

If I compare the "original" `targets` array of "browsers that have support for CSS variables" with adjusted `targets` array of "browsers we can support without polyfills", potentially we _could_ support, but _are not yet_ compatible with:  
- Chrome 49 to 52
- Edge 16 to 18
- Firefox 31 to 49
- iOS Safari 9.3 to 10.3
- Opera 36 to 39
- Safari 9.1 to 10.1
- Samsung 5

Of course these browsers are the ones where most errors will occur, they are old but still account for **2%** of today's browsers according to [this Browserslist query](https://browserslist.dev/?q=Q2hyb21lIDQ5LTUyLCBFZGdlIDE2LTE4LCBGaXJlZm94IDMxLTQ5LCBpT1MgOS4zLTEwLjMsIE9wZXJhIDM2LTM5LCBTYWZhcmkgOS4xLTEwLjEsIFNhbXN1bmcgNQ%3D%3D) (in my case, we're talking thousands of users a day).

Missing polyfills is one concern that we can solve with the rewrite but I don't think this is really an issue as `core-js` add polyfills to fully comply with standards, although Linaria code most probably runs fine anyway ("partial" compliance is good enough).  
The main concern (the one that brought me here) is the amount of `SyntaxError` due to modern JS syntax. This PR brought more transforms but some are still missing, for example:
- `SyntaxError: Unexpected token =>`: **transform-arrow-functions**
- `SyntaxError: Unexpected keyword 'const'`: **transform-block-scoped-functions**

We'll have those (and some others as equally useful), after we specify the browsers' versions in the `targets` and rewrite JS to avoid extra `core-js` :)

Issue #704, opened 2020-11-23
@linaria/babel doesn't work


ym-project:
## Environment

**packages**
* "@babel/cli": "^7.12.7"
* "@babel/core": "^7.12.7"
* "@linaria/babel": "^3.0.0-beta.0"
* "@linaria/webpack4-loader": "^3.0.0-beta.0"
* "babel-loader": "^8.2.1"
* "webpack": "^4.44.2"
* "webpack-cli": "^4.2.0"

**software**
* nodejs: v15.2.1
* os: ArchLinux 5.5.13-arch2-1

## Description

When I try to bundle the project I get an error:
```sh
ERROR in ./src/index.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Cannot find module '@linaria/babel-preset-babel' from '.../linaria-test'
- If you want to resolve "@linaria/babel", use "module:@linaria/babel"
    at Function.resolveSync [as sync] (.../linaria-test/node_modules/resolve/lib/sync.js:90:15)
    at resolveStandardizedName (.../linaria-test/node_modules/@babel/core/lib/config/files/plugins.js:101:31)
    at resolvePreset (.../linaria-test/node_modules/@babel/core/lib/config/files/plugins.js:58:10)
    at loadPreset (.../linaria-test/node_modules/@babel/core/lib/config/files/plugins.js:77:20)
    at createDescriptor (.../linaria-test/node_modules/@babel/core/lib/config/config-descriptors.js:154:9)
    at .../linaria-test/node_modules/@babel/core/lib/config/config-descriptors.js:109:50
    at Array.map (<anonymous>)
    at createDescriptors (.../linaria-test/node_modules/@babel/core/lib/config/config-descriptors.js:109:29)
    at createPresetDescriptors (.../linaria-test/node_modules/@babel/core/lib/config/config-descriptors.js:101:10)
    at presets (.../linaria-test/node_modules/@babel/core/lib/config/config-descriptors.js:47:19)
```
As I understand babel can not to find `@linaria/babel`. Instead of necessary package babel try to find  `@linaria/babel-preset-babel`. Previous version (2.0) has `linaria/babel` package and it works great.

## Reproducible Demo
https://github.com/ym-project/linaria-babel-bug

Anber:
Hi @ym-project 

I've also faced that problem. As a workaround, you can replace `@linaria/babel` in a config with `module:@linaria/babel`.

ym-project:
> Hi @ym-project
[...]

PR #720, opened 2020-12-08
fix(babel): rename @linaria/babel to @linaria/babel-preset (#704)

Anber:
## Motivation

Babel forces to use standardized names for plugins and presets.

## Summary

`@linaria/babel` has been renamed `@linaria/babel-preset`, so that Babel will allow to use it without prefix `module:` in a list of presets. (#704)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,13 @@ Zero-runtime CSS in JS library.
 ## Installation
 
 ```sh
-npm install @linaria/core @linaria/react @linaria/babel @linaria/shaker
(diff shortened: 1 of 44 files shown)


thymikee (approved):
👍 This is a breaking change, so worth bumping major version when releasing

Anber:
@thymikee it's BC if we compare it with the first beta. I think it's ok to break something between `beta.0` and `beta.1`. The old `linara/babel` will continue to work, so it's not even BC between 2.0 and 3.0.

thymikee:
Okies

PR #725, opened 2020-12-31
docs: fixing links to other doc files

joeycozza:
## Motivation

Links in some doc files were pointing to wrong location with relative path.

## Summary
Tweaked the relative path to correctly point to the doc file


jayu (approved):
Thanks!

PR #728, opened 2021-01-11
fix: loadOptions test regex

### LINKED ISSUE #727: Evaluator behaves differently on nix and win OS due to weak path matching
## Environment

- Linaria version: 3.0.0-beta.1
- Bundler (+ version): Webpack 4.4.2
- Node.js version: 12 and higher
- OS: Windows, macOS, any other nix-based

## Description

Evaluator throws `_react.memo is not a function` error on windows if one try to wrap react memo component into styled. Essentially this happens because of loadOption's regex for matching `node_modules` in file path whic works properly in nix env: https://github.com/callstack/linaria/blob/d0430f3b3d670d3f837b108c2864539734497360/packages/babel/src/utils/loadOptions.ts#L31

## Reproducible Demo
Just make wrap any `memo`ized react component into linaria's `styled` fn.

Nedgeva:
## Motivation

Fixes #727 

## Summary

Changes test regex in loadOptions to match `node_modules` path on Win machines, not only nix.

## Test plan

1. create sample project with react, linaria and webpack
2. make dummy react component wrapped into react's `memo`
3. wrap abovementioned react component into linaria's `styled`
4. run build in different envs (e.g. linux, macos, windows)
5. build should finish with success

### DIFF
--- a/packages/babel/src/utils/loadOptions.ts
+++ b/packages/babel/src/utils/loadOptions.ts
@@ -28,7 +28,7 @@ export default function loadOptions(
       },
       {
         // The old `ignore` option is used as a default value for `ignore` rule.
-        test: ignore ?? /\/node_modules\//,
+        test: ignore ?? /[\\/]node_modules[\\/]/,
         action: 'ignore',
       },
     ],


Anber (approved):

Anber:
Thank you @Nedgeva!

Issue #731, opened 2021-01-30
@linaria/cli use invalid bin field


huang-xiao-jian:
## Environment

- Linaria version: 3.0.0-beta.1
- Bundler (+ version):
- Node.js version:
- OS:

## Description

`@linaria/cli` use invalid `bin` field value, which generate binary link as `cli`

![image](https://user-images.githubusercontent.com/4002210/106361655-0bbbde00-635a-11eb-8067-9c2060f46af9.png)

use below instead:

```json
{
  "bin": {
     "linaria": "bin/linaria.js"
  }
}
```

Issue #732, opened 2021-01-31
@linaria/cli should resolve output filename relative to source root


huang-xiao-jian:
## Environment

- Linaria version: 3.0.0-beta.1
- Bundler (+ version): 
- Node.js version:
- OS:

## Description

before linaria extract:

![image](https://user-images.githubusercontent.com/4002210/106378983-b0363280-63e3-11eb-8bc8-3373f8cd45f9.png)

execute command:

```shell
yarn cli 'src/**/*.ts' -o cjs -r src
```

after linaria extract:

![image](https://user-images.githubusercontent.com/4002210/106379008-ebd0fc80-63e3-11eb-944f-e0a24968d30e.png)

expected output:

![image](https://user-images.githubusercontent.com/4002210/106379043-3b172d00-63e4-11eb-98f7-e797c2640710.png)

## Reproducible Demo

https://repl.it/@bornkiller/TrustworthyDisloyalDribbleware

PR #733, opened 2021-01-31
fix: resolve output filename relative to source root, use valid bin field

huang-xiao-jian:
## Summary

resolve issue #731 and #732 

## Test plan

see detail in issue #731 and #732


Anber (approved):

Anber:
@huang-xiao-jian thank you!

PR #763, opened 2021-05-28
fix: @linaria/rollup compliant vite

wmzy:
## Motivation

@linaria/rollup inject the css file with a real file path, but the css flie not exist in file system.
And vite [rewrite the absolute path with a relative path](https://github.com/vitejs/vite/blob/5745a2e8072cb92d647662dc387e7f12b2841cab/packages/vite/src/node/plugins/importAnalysis.ts#L195), so vite server could not find the the css file from file system and @linaria/rollup could not find it from `cssLookup` object.

I add a `@linaria:` prefix to css filename so that rollup and vite see it as a **virtrul** file.

## Summary

Make it work for vite.

### DIFF
--- a/packages/rollup/src/index.ts
+++ b/packages/rollup/src/index.ts
@@ -49,7 +49,7 @@ export default function linaria({
       let { cssText } = result;
 
       const slug = slugify(cssText);
-      const filename = `${id.replace(/\.js$/, '')}_${slug}.css`;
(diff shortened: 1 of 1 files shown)


Anber (approved):

Anber:
Thank you @wmzy!

Anber:
Hi @wmzy,

There is a report about regression after this PR https://github.com/callstack/linaria/issues/778

Issue #813, opened 2021-08-06
@linaria/babel-preset doesn't have @linaria/shaker as dependency but requires in `loadOptions.js`


soldonii:
## Environment

I'm setting up linaria through babel and rollup.
As docs described, I installed `@linaria/babel-preset` and add it to `.babelrc` file.

![image](https://user-images.githubusercontent.com/49357004/128469740-e6d2fc3b-2280-4c2b-8aaf-d1f04675aef1.png)

and it throws error when bundling via rollup. Below is error message.

![image](https://user-images.githubusercontent.com/49357004/128470766-c50da88a-9554-4b1c-a287-7a123c268986.png)

The problem is, `@linaria/babel-preset` doesn't have `@linaria/shaker` as dependency, but [it requires inside of `loadOptions.js` in `@linarai/babel-preset` package](https://github.com/callstack/linaria/blob/master/packages/babel/src/utils/loadOptions.ts#L25-L27).

Below is installed package when I type `yarn add @linaria/babel-preset`. It doesn't seem to include `@linaria/shaker`.
![image](https://user-images.githubusercontent.com/49357004/128470082-2b8311fb-51c9-4d34-bfff-07c9a88a0461.png)

Is there anything that I've done wrong?

- Linaria version: latest(@linaria/babel-preset: "^3.0.0-beta.4")
- Bundler (+ version): rollup(2.3.0)
- Node.js version: 14.16.0
- OS: mac

soldonii:
it happens whether I install `@linaria/shaker` or not.
I installed every package mentioned in docs(`@linaria/core @linaria/react @linaria/babel-preset @linaria/shaker`)
Also `yarn add --dev rollup-plugin-css-only @linaria/rollup`

Anber:
Hi @soldonii!
Do you use yarn pnp?

soldonii:
> Hi @soldonii!
> Do you use yarn pnp?

yes I do. Is it related to version of yarn? :)

PR #819, opened 2021-08-14
feat: make @linaria/shaker optional

kazuma1989:
## Motivation

I want to use Linaria without evaluating the interpolated values. So I don't need to install `@linaria/shaker` but `@linaria/babel-preset` requires the package if no evaluator rules are active.

The default evaluator rules will be skipped when any rules option is set. It makes `@linaria/shaker` optional.

Related #813.

## Test plan

A rollup example. This should work without `@linaria/shaker` installed.

```js
import linaria from '@linaria/rollup';

export default {
  plugins: [
    linaria({
      rules: [],
    }),
  ],
};
```


Anber (approved):

Anber:
Thank you @kazuma1989!

steffenmllr:
@Anber any chance this get's published on npm?

PR #825, opened 2021-08-28
feat: Add {file, name, dir, ext} to `classNameSlug` resolution

### LINKED ISSUE #571: Add a third argument to the preprocessor option that gives us the file name

### LINKED ISSUE #650: Pass filename to classNameSlug function?
## Describe the enhancement

It would be helpful to have the .js filename passed to the classNameSlug function (if a user uses the function approach in their config rather than the string approach).  This gives the user additional options for configuring the classname returned from the function.

## Motivation

I've been happily using CSS Modules (via the babel plugin) for the last few years with React.  I'm always trying to improve the developer experience (DX) for my team, and folks find it irritating to have the CSS in a separate file (and then, being lazy, they often revert to using inline styles and never get around to moving those into the associated css file).  So, I'm looking at various options for having the CSS in the file with the JSX, including linaria.  I'm really liking linaria thus far (started today) for it's simplicity, ease of customization, and straightforward nature compared to many of the other css-in-js libraries.

My #1 approach towards improving DX has been to keep the mental model simple.  This makes easy things easy and hard things not as hard, so folks are more willing to go the extra mile.  This has paid great benefits for our team, and things have been kind of fun.

One of the key mechanisms I've used is to have the base className of any component be the name of the filename in which the component exists (or the identically named .scss file that sits next to it).  We have a largish project, and no one on the team is familiar with the all of it, but a dev who has a feature to work on (or when troubleshooting), can open the app in a browser, inspect the HTML, and see something like this:

```html
<div class="app___">
    <div class="bp3-navbar app-header___">...</div>
    <div class="app-body___">...</div>
</div>
```

From inspecting the HTML, the dev can easily get an idea of the component structure and can know exactly which file in the project they need to open for the component(s) that they want to look at.  The filenames are actually taken from .scss files, but these files are always right next to the .tsx file to which that CSS file applies (we use Typescript).  We also append a short hash (couple of characters) to the classnames, which I'm not showing above.

What we do is, using CSS Modules, the component itself has a "styleName" of "_" (underscore), and then any children rendered directly by that component will be given local-scoped classNames using a pseudo-BEM style in the format: <scss filename>__<local classname>.  This also makes it super easy to zero in on where the CSS comes from that is driving the style of each component or it's directly-rendered children.

Basically I'm trying to maintain something like this with in-file CSS solutions, including linaria, and after poking around for a little bit, it looks pretty easy to accomplish.

Below is are the app-header.scss and app-header.tsx from a small demo app that I'm messing with, in case that helps explain things.

**app-header.scss:**
```typescript
._ {
    height: 48px;
    display: flex;
    justify-content: space-between;
    user-select: none;
    cursor: default;
}

.logo {
    width: 38px;
    src: url("/images/logo-red.png");
}

.brand {
    display: flex;
}

.title {
    font-size: 22px;
    color: #9292a1;
}

.tabs {
    display: inline-flex !important;
    font-weight: bold;

    /* make the active tab indicator a little taller and move it down a little */
    :global {
        .bp3-tab-list.bp3-large {
            .bp3-tab-indicator {
                bottom: -2px !important;
                height: 5px !important;
            }
        }
    }
}

.tab {
    font-size: 14px;
}

.name {
    margin-left: 5px;
    font-size: 16px;
    color: #4e4e61;
}

```

**app-header.tsx:**
```typescript
import React from "react";
import { Navbar, Text, Tabs } from "@blueprintjs/core";
import Avatar from "components/avatar";

import css from "./app-header.scss";

const AppHeader = () => (
    <Navbar styleName="_">
        <Navbar.Group>
            <img styleName="logo" src="/logo.png" alt="" />
            <Navbar.Divider />
            <Text className={css.title}>Demo App</Text>
        </Navbar.Group>

        <Navbar.Group className={css.tabs}>
            <Tabs id="main" large>
                <Tabs.Tab id="analytics">Analytics</Tabs.Tab>
                <Tabs.Tab id="worklists">Worklists</Tabs.Tab>
            </Tabs>
        </Navbar.Group>

        <Navbar.Group>
            <Avatar size="40px" />
            <Text className={css.name}>First Lastname</Text>
        </Navbar.Group>
    </Navbar>
);

export default AppHeader;
```

## Possible implementations

It's really easy to implement:  in GenerateClassNames.ts, where classNameSlug is called if it is a function, just change it to something like:
```typescript
if (typeof options.classNameSlug === 'function') {
    try {
      const fn = state.file.opts.filename;
      const basename = _path.basename(fn, _path.extname(fn));
      className = toValidCSSIdentifier(options.classNameSlug(slug, displayName, basename ));
    } catch {
      throw new Error(`classNameSlug option must return a string`);
    }
  }
```

I tried this out on a fork, and it works nicely.  The other thing I had to do was set the extension to ".css" in the linaria/loader config so that linaria didn't generate names like "app-header.linaria.css" but instead just "app-header.css".  Then the filename is used in the classname with a custom classNameSlug function, and I can get the exact same output as with CSS Modules.  In playing around this evening, I've converted some of the components to linaria, but the rest are still using CSS Modules.  Both are playing well together though loaded via different rules in the webpack.config, and linaria is playing well with Typescript via having ts-loader in the chain before linaria/loader.  

## Related Issues

I'm not aware of any, just started looking at linaria today.  So far, so good.

mikestopcontinues:
Fixes #650
Fixes #571

## Motivation

If you want to know where a particular class is coming from, you need more than a hash or even the original title of the class. How many `.inner` or `.title` classes have you seen in a large codebase, right?

## Summary

This PR adds four new vars to `classNameSlug` resolution. It also adds the original two `hash` and `title` to the object, with the idea that a future breaking change might sandwich the function parameters down to just this one.

While just `file` would have been sufficient for function style, I added `dir, name, ext` for ease of use with string-style. Here's what you would expect in each:

```ts
{
  hash: 'qdegrke',
  title: 'displayName',
  file: 'this/is/your/module/Component/SubComp.tsx',
  name: 'SubComp',
  ext: 'tsx',
  dir: 'Component',
}
```

## Test plan

Added tests.

### DIFF
--- a/packages/babel/__tests__/__snapshots__/babel.test.ts.snap
+++ b/packages/babel/__tests__/__snapshots__/babel.test.ts.snap
@@ -97,7 +97,7 @@ exports[`handles fn passed in as classNameSlug 1`] = `
 "import { styled } from '@linaria/react';
 export const Title = /*#__PURE__*/styled('h1')({
   name: \\"Title\\",
-  class: \\"t17gu1mi_42_Title\\"
+  class: \\"t17gu1mi_Title_t17gu1mi_Title__tests___app_index_js_index__js_app\\"
 });"
 `;
 
@@ -105,7 +105,7 @@ exports[`handles fn passed in as classNameSlug 2`] = `
 
 CSS:
 
-.t17gu1mi_42_Title {
+.t17gu1mi_Title_t17gu1mi_Title__tests___app_index_js_index__js_app {
   font-size: 14px;
 }
 
@@ -389,6 +389,26 @@ Dependencies: NA
 
 `;
 
+exports[`removes fake replacement patterns in string classNameSlug 1`] = `
+"import { styled } from '@linaria/react';
+export const Title = /*#__PURE__*/styled('h1')({
+  name: \\"Title\\",
+  class: \\"__\\"
+});"
+`;
+
+exports[`removes fake replacement patterns in string classNameSlug 2`] = `
+
+CSS:
+
+.__ {
+  font-size: 14px;
+}
+
+Dependencies: NA
+
+`;
+
 exports[`replaces unknown expressions with CSS custom properties 1`] = `
 "import { styled } from '@linaria/react';
 export const Title = /*#__PURE__*/styled(\\"h1\\")({
@@ -552,7 +572,7 @@ exports[`uses string passed in as classNameSlug 1`] = `
 "import { styled } from '@linaria/react';
 export const Title = /*#__PURE__*/styled('h1')({
   name: \\"Title\\",
-  class: \\"testSlug\\"
+  class: \\"t17gu1mi_Title__tests___app_index_js_index__js_app\\"
 });"
 `;
 
@@ -560,7 +580,7 @@ exports[`uses string passed in as classNameSlug 2`] = `
 
 CSS:
 
-.testSlug {
+.t17gu1mi_Title__tests___app_index_js_index__js_app {
   font-size: 14px;
 }
 
--- a/packages/babel/__tests__/babel.test.ts
+++ b/packages/babel/__tests__/babel.test.ts
@@ -43,7 +43,29 @@ it('uses string passed in as classNameSlug', async () => {
       font-size: 14px;
     \`;
 `,
-    { classNameSlug: 'testSlug' }
+    {
+      classNameSlug: ['hash', 'title', 'file', 'name', 'ext', 'dir']
+        .map((s) => `[${s}]`)
+        .join('_'),
+    }
+  );
+
+  expect(code).toMatchSnapshot();
+  expect(metadata).toMatchSnapshot();
+});
+
+it('removes fake replacement patterns in string classNameSlug', async () => {
+  const { code, metadata } = await transpile(
+    dedent`
+    import { styled } from '@linaria/react';
+
+    export const Title = styled('h1')\`
+      font-size: 14px;
+    \`;
+`,
+    {
+      classNameSlug: `[not]_[actual]_[replacements]`,
+    }
   );
 
   expect(code).toMatchSnapshot();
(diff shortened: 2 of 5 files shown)


mikestopcontinues:
I don't think the circleci issue has to do with my code. I couldn't get it to run properly on my machine either, but `yarn watch` and `yarn test` both work.

Anber:
Hi @mikestopcontinues! Thank you for your PR.
> I don't think the circleci issue has to do with my code. I couldn't get it to run properly on my machine either, but yarn watch and yarn test both work.

There is a problem with types in new code:
```
src/visitors/GenerateClassNames.ts(150,59): error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'ClassNameSlugVars'.
  No index signature with a parameter of type 'string' was found on type 'ClassNameSlugVars'.
```
You can either relax the definition of `ClassNameSlugVars` by changing it to `Record<string, string>` or replace `v in slugVars` with some more strict type guard (like `const isSlugVar = (v: string): v is keyof ClassNameSlugVars => v in slugVars`).

All checks can be run by `yarn prepare`

mikestopcontinues:
@Anber Sorry about that. I just assumed what the circleci issue was. I added one additional test for non-existing replacement patterns to make sure the logic remained consistent.

Anber (approved):

PR #827, opened 2021-08-31
Fixes for `--exactOptionalPropertyTypes` TS flag

geakstr:
TypeScript 4.4 introduced [new `--exactOptionalPropertyTypes` flag](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#exact-optional-property-types---exactoptionalpropertytypes). With this flag enabled typescript compiler strictly distinguish between `?` and `undefined` type. Currently [`@types/react:17.0`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c134157852d95b32a5feeb675e13489b76fbeda0/types/react/index.d.ts#L1831-L1842) package has `className?: string | undefined` and `style?: CSSProperties | undefined` specified for html attributes, but linaria types are only marked as `?` but `undefined`. This patch fixes this incompatibility

### DIFF
--- a/packages/react/src/styled.ts
+++ b/packages/react/src/styled.ts
@@ -63,13 +63,15 @@ interface IProps {
 // If styled wraps custom component, that component should have className property
 function styled<TConstructor extends React.FunctionComponent<any>>(
   tag: TConstructor extends React.FunctionComponent<infer T>
-    ? T extends { className?: string }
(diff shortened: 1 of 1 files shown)


Anber (approved):

Anber:
Thank you @geakstr!

PR #846, opened 2021-09-26
Remove "hmr" option from Webpack docs

Daniel15:
After upgrading Webpack, I was getting this error:

```
ValidationError: Invalid options object. Mini CSS Extract Plugin Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'hmr'. These properties are valid:
   object { publicPath?, emit?, esModule?, layer? }
```

Seems like this option isn't supported any more.


Anber (approved):

Anber:
Thank you @Daniel15!

PR #862, opened 2021-11-04
Update BASICS.md

isumix:
Should probably be without quotes here as it is compiled CSS


Anber:
Hi @isumix 
Yep, it definitely should be without quotes. Thank you!

Anber (approved):

PR #830, opened 2021-09-02
fix(webpack): pass all user resolve options to loader (fixes #658)

5angel:
## Motivation

This is a fix for #658 and other related bugs

## Summary

Now all options are passed down to loader.

PR #874, opened 2021-11-28
fix(webpack): better merge for configs and fallback for async plugins

### LINKED ISSUE #855: Webpack error with imported variables
## Environment

- Linaria version: `3.0.0-beta.13`
- Bundler (+ version): `webpack@5.59.0`
- Node.js version: `14.15.1`
- OS: `macOS@11.6`

## Description

We've been using Linaria successfully in our projects for a while now. However, I have encountered a strange issue recently with using the Webpack loader, but which works with CLI.

If I use a constant that is imported from another module, Webpack blows up with `finishWithoutResolve` error.

For example:
```js
import { styled } from '@linaria/react';
import { device, size } from './breakpoints';

const Description = styled.p`
  background-color: orange;
 
  @media ${ device.tablet } {    
    background-color: blue;
  }
`;
```

Fails to build, erroring with:
`Can't resolve './breakpoints' in '/path/to/linaria-webpack-bug/src'`

## Reproducible Demo

A reproducible repo with steps is available here:
https://github.com/kinetifex/linaria-webpack-bug

Anber:
## Motivation

Fixes #855 and regression from #830

## Summary

In #830 we decided to use the whole Webpack configuration in Linaria's resolver. It appeared, that it wasn't the wisest decision because it introduced a lot of breaking changes. This PR introduces a better merge for Webpack config with default Linaria's config and adds a fallback if Webpack configuration breaks the resolver by using some async plugins.

### DIFF
--- a/packages/webpack5-loader/src/index.ts
+++ b/packages/webpack5-loader/src/index.ts
@@ -14,7 +14,7 @@ import findYarnWorkspaceRoot from 'find-yarn-workspace-root';
 import type { RawSourceMap } from 'source-map';
 import cosmiconfig from 'cosmiconfig';
 import { EvalCache, Module, transform } from '@linaria/babel-preset';
-import { debug } from '@linaria/logger';
+import { debug, notify } from '@linaria/logger';
 
 const workspaceRoot = findYarnWorkspaceRoot();
 const lernaConfig = cosmiconfig('lerna', {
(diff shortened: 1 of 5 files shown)


packages/webpack5-loader/src/index.ts
```diff
@@ -56,18 +61,47 @@ export default function webpack5Loader(
...
-    // There is this.resolve, but it's asynchronous
-    // Another option is to read the webpack.config.js, but it won't work for programmatic usage
-    // This API is used by many loaders/plugins, so hope we're safe for a while
-    {
-      ...(this._compilation?.options.resolve ?? {}),
+  // this._compilation is a deprecated API
+  // However there seems to be no other way to access webpack's resolver
+  // There is this.resolve, but it's asynchronous
+  // Another option is to read the webpack.config.js, but it won't work for programmatic usage
+  // This API is used by many loaders/plugins, so hope we're safe for a while
```
ntucker, line 68: this._compilation is undefined in the latest webpack

ntucker: Does https://webpack.js.org/api/loaders/#thisgetresolve work?

Issue #851, opened 2021-10-05
Allow re-exporting of templates such as `styled` and `css` from internal files


jpnelson:
## Describe the feature

We'd like to be able to re-export the `css` and `styled` template literals. Currently, the babel plugin will only detect and replace imports from the linaria library. This, for example, would not work:

`./my-folder/linaria`

```js
export { css } from 'linaria';
```

`./index.js`

```js
import { css } from './my-folder/linaria'

const className = css`
  background: red;
`
```

Since `css` is not being imported directly from `linaria`

## Motivation

We have some linaria specific utilities (eg. shared styles and the like) that we house in a separate folder. It's a little cumbersome to also import the public API – if we could, we would re-export it from our internal folder to keep everything in one place.

## Possible implementations

We could implement this either by adding logic to try to resolve the imports (might be tricky with custom configs) or to allow specifying which import paths may contain `styled` and `css` – I'd be happy to take direction on which of these is preferred.

Anber:
Hi @jpnelson 

Just custom import paths will not help because `css` and `styled` can be reexported in different ways and with different names. Probably we need to introduce a new config option for a custom resolve function and completely refactored `hasImport` which will use that option.

jpnelson:
@Anber I had a go at this here: https://github.com/callstack/linaria/pull/882 – it turned out to be not too tricky, let me know what you think! 🙏

jpnelson:
Done in https://github.com/callstack/linaria/pull/882

PR #882, opened 2021-12-06
feat(resolver): add custom resolver option to support re-exporting of linaria libs

jpnelson:
## Motivation

This addresses https://github.com/callstack/linaria/issues/851

## Summary

The docs should explain the solution here fairly well – let me know if more detail would help.

Basically, adds a custom resolver that can be used in `hasImport` to allow building linaria libs from different folders

## Test plan

I tested this by trying it out in the `website/` folder. It worked there, and I added some snapshot tests too


Anber (approved):

Anber:
Thank you @jpnelson!

Issue #881, opened 2021-12-03
In-memory cache isn't compatible with thread-loader


malash:
## Environment

- Linaria version: 2.2.0
- Bundler (+ version): Webpack 5.64.4
- Node.js version: 16.3.0
- OS: macOS 12.0.1

## Description

The PR https://github.com/callstack/linaria/pull/879 introduced an in-memory cache to fix rebuild issue and accelerate bundle time. It's awesome, but when I try to integrate the latest version of Linaria in my project, I found some output CSS code is missing.

I deep dived and realize the in-memory cache may not be compatible with [thread-loader](https://webpack.js.org/loaders/thread-loader/). `thread-loader` use a thread poll to execute loaders in non-main threads, which can hugely speed up Webpack's building time in a large project. Since the threads in NodeJS cannot share the same closure context, the [cssLookup](https://github.com/callstack/linaria/pull/879/files#diff-d279ac35f28bc8fef0b98f559d88b3169b2709c3c289b7de384d72f4f73b1dbd) map could be empty when running `outputCssLoader.ts` in a brand new thread.

I'm not sure is this a bug or a feature request? because not all Webpack projects use the `thread-loader`. But in our project, the build time would become non-acceptable if I disabled the `thread-loader`. So I can provide a posable solution if you agree to fix this issue:

## Solution

Add a new options to return the right of choice to Linaria users.

- `cacheType: 'memory' | 'filesystem'` (default: `'memory'`):

This option is inspired from Webpack's [cache.type](https://webpack.js.org/configuration/cache/#cachetype). If set to `memory` the loader works in the same way as 2.2.0, and if set to `filesystem` it works in the same way as 2.1.0.

- `cacheDirectory: string` (default: `'.linaria-cache'`):

Same as previous. Only works when `cacheType: 'filesystem'`

## Reproducible Demo

https://github.com/malash/linaria-issue-881

You can enable/disable [this line](https://github.com/malash/linaria-issue-881/blob/master/webpack.config.js#L26-L29) and run `yarn webpack`, the check the result of `dist/styles.css`.

![image](https://user-images.githubusercontent.com/1812118/144583450-5aaf732b-20a1-44e3-9a8d-29bbc03347f7.png)

malash:
cc @Anber

Anber:
Hmm…
thread-loader docs say, that loaders in threads have some limitations:
- Loaders cannot emit files.
- Loaders cannot use custom loader API (i. e. by plugins).
[...]

PR #884, opened 2021-12-13
fix(webpack): add cacheProvider to support customized cache

malash:
## Motivation

This PR try to fix https://github.com/callstack/linaria/issues/881 . This PR is mainly inspired from https://github.com/callstack/linaria/commit/a281440271cb29d4d2e2b0cb3b2fee26f1e4bc69

## Summary

This PR add a new option `cacheProvider` for Linaria loader.

- `cacheProvider: undefined | string | ICache` (default: `undefined`):
  By default Linaria use a memory cache to store temporary CSS files. But if you are using this loader with [thread-loader](https://www.npmjs.com/package/thread-loader) you should use some consistent cache to prevent [some unexpected issues](https://github.com/callstack/linaria/issues/881). This options support a `ICache` instance or a path to NodeJS module which export a `ICache` instance as `module.exports`
  > ```
  > interface ICache {
  >   get: (key: string) => Promise<string>;
  >   set: (key: string, value: string) => Promise<void>
  > }
  > ```

## Test plan

I run `yarn build` in this repo and copied `libs` folder to my Linaria project's `node_modules/linaria` folder.

### DIFF
--- a/docs/BUNDLERS_INTEGRATION.md
+++ b/docs/BUNDLERS_INTEGRATION.md
@@ -1,6 +1,7 @@
 # Bundlers Integration
 
 ## Jump To
+
 - [webpack](#webpack)
 - [Rollup](#Rollup)
 - [Svelte](#Svelte)
@@ -11,7 +12,7 @@ If you use Babel in your project, make sure to have a [config file for Babel](ht
 
 ## Bundlers
 
-Please note, that `@babel/core` is a peer dependency of all loaders. Do not forget to add it to `devDependencies` list in your project. 
+Please note, that `@babel/core` is a peer dependency of all loaders. Do not forget to add it to `devDependencies` list in your project.
 
 ### webpack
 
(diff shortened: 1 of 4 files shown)


Anber:
Hi @malash!
Thank you for the pull request!

By pulling back filesystem-based cache we are pulling back all the problems with detecting the project root and the cache directory. I'm thinking… maybe we can just introduce a `cache` option, instead of `cacheType`? It can be an implementation of a simple interface like `interface ICache { get: (key: string) => string; set: (key: string) => string  }`. The default implementation will be an empty `Map`, but it can be replaced with whatever you want: mock, filesystem or even Redis (if someone wants to use distributed build :)). What do you think?

malash:
@Anber Good idea. 

I think the interface should be async. Because async `fs` provide better performance than sync `fs`, and Redis might only support async API. And to implement the feature, the Webpack loader should be refactored with [async loader](https://webpack.js.org/api/loaders/#asynchronous-loaders).

```tsx
interface ICache { 
  get: (key: string) => Promise<string>;
  set: (key: string, value: string) => Promise<void>
}.
```

If ok, I can refactor the PR.

malash:
@Anber I realized there might be another problem if we chose the `ICache` plan. 

In my case I want to use Linaria with `thread-loader`, and `thread-loader` always [serialize the loader context](https://github.com/webpack-contrib/thread-loader/blob/2e1dfff9fc7133c741ffdd4f1da5a7693b637ad0/src/WorkerPool.js#L109-L115) for cross-thread communication. Unfortunately `function`s cannot be serialized, so this plan still cannot solve the problem https://github.com/callstack/linaria/issues/881 .

How about implementing the `cache` option like this:

```tsx
cache: 'memory' | 'filesystem' | ICache;
```

It should be scalable enough ( even someone want to use Redis-based distributed build ).

Anber:
> cache: 'memory' | 'filesystem' | ICache;

It still requires `find-yarn-workspace-root` and other hacks…
Maybe instead of cache implementation, we can specify a path to a file/module with implementation? It would be serializable.

malash:
@Anber I refactored the PR, could you please take a look?

I tested this branch in [my project](https://github.com/airbnb/goji-js/pull/145/files#diff-ae827d34af4c1066afb39b000bc21a24a621798a4c1049e8e466c89427b5568c) and it works.

Anber:
@malash thank you for your awesome work!

Anber (approved):

Anber:
I'll publish it as 2.3 later today.

Anber:
Released

PR #886, opened 2021-12-14
feat(types): add style type declaration

bobsam:


packages/react/src/styled.ts
```diff
@@ -69,7 +69,7 @@ interface IProps {
 // If styled wraps custom component, that component should have className property
 function styled<TConstructor extends React.ComponentType<any>>(
   tag: TConstructor extends React.ComponentType<infer T>
-    ? [T] extends [{ className?: string | undefined }]
+    ? [T] extends [{ className?: string | undefined; style?: React.CSSProperties | null; }]
```
Anber, line 72: It isn't needed here, since `style` is required only if props-based interpolation is used and for such case, we already have `style` presence check (line 208).

PR #889, opened 2021-12-18
fix(webpack): add cacheProvider for Linaria v3

malash:
## Motivation

This PR add the same feature as https://github.com/callstack/linaria/pull/884 for Linaria v3.

## Summary

See https://github.com/callstack/linaria/pull/884

## Test plan

I tested this PR in https://github.com/malash/linaria-issue-881/commit/ae8ecca1d8adb812c39ede9c897d405549fa5ae4 with `cacheProvider: undefined` and `cacheProvider: require.resolve("./linariaFileCache")` and they work as expected.

@Anber

PR #890, opened 2021-12-18
fix(webpack): fix error when cacheProvider is undefined

malash:
## Motivation

While cherry picking `cacheProvider` to Linaria v3 I found a same bug in https://github.com/callstack/linaria/pull/884

## Summary

If `cacheProvider` is not set ( `undefined` ), the query param `cacheProvider=${encodeURIComponent(cacheProvider)}`  became a string `'undefined'` and result in a runtime error `Cannot find module 'undefined'`.

## Test plan

I'v test `cacheProvider: undefined` in https://github.com/callstack/linaria/pull/889

### DIFF
--- a/src/loader.ts
+++ b/src/loader.ts
@@ -109,7 +109,7 @@ export default function loader(
       .then((cacheInstance) => cacheInstance.set(this.resourcePath, cssText))
       .then(() => {
         const request = `${outputFilename}!=!${outputCssLoader}?cacheProvider=${encodeURIComponent(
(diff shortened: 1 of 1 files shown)


malash:
@Anber Could you mind public a new version of Linaria v2, since this bug would affect most of normal Linaria users.

Anber:
Hi @malash 
I would like to, but there is a big issue #891 introduced in #879. I have to fix it first.

Issue #912, opened 2022-01-26
(atomic): string serialization for atomic styles produced


jpnelson:
## Describe the enhancement

Currently, atomic styles produces an object in the resulting code like this:

```ts
const atoms = css`
  background: red;
  width: 100%;
`;

// gets compiled to
const atoms = {
  background: 'atm_foo',
  width: 'atm_bar',
}
```

I would propose to change it to this:

```ts
const atoms = css`
  background: red;
  width: 100%;
`;

// gets compiled to
const atoms = 'atm_hash1-foo atm_hash2-bar';
```

Where `hash1` is a hash of `background` and `hash2` is a hash of `width`.

`cx` can then be modified to do something like this:

```ts
const KEY_VALUE_SEPARATOR = '-';

const cx = function cx() {
  const presentClassNames = Array.prototype.slice
    .call(arguments)
    .filter(Boolean);

  // `cx` is passed all strings with encoded key:value, we need to dedupe
  const map = {};

  for (let str of presentClassNames) {
    // className could be the output of a previous cx call, so split by ' ' first
    const individualClassNames = str.split(' ')
    
    for (let className of individualClassNames) {
    	const [key, value] = className.split(KEY_VALUE_SEPARATOR);
    	map[key] = `${key + KEY_VALUE_SEPARATOR + value}`;
    }
  }

  return Object.values(map);
};
```

## Motivation
The issue we have today is people using `cx` multiple times, and not getting correct merging behavior. Generally this happens because people merge styles before passing them around. The issue is that effectively,

```
cx(a, cx(b, c)) !== cx(a, b, c)
```

## Possible implementations

From benchmark testing, doing string manipulation isn't much worse than merging objects. However, if anyone felt strongly about it, we could:

1. Provide an option to choose between the two representations
2. Provide a different `cx` util

I don't think that extra complexity is needed though, so I'd recommend

yume-chan:
Is this issue already fixed and can be closed?

espretto:
Yes, looking at #934 I think so, too. In this small issue this fix seems so underrated. @jpnelson Is there anything to add here or is this done?

Issue #902, opened 2022-01-13
Use inline-style-expand-shorthand 


jpnelson:
## Describe the enhancement

To correctly merge together atomic styles, we should expand shorthand properties. https://github.com/johanholmerin/style9/blob/10d282d17732a15409fcdc59ba75ed152f219e3f/docs/Usage-guide.md#shorthands contains some info on how this can work. We just need to explode properties like `padding: 5px` into four atoms: one for top, right, bottom, left.

## Motivation

This allows us to correctly override one of padding values

## Possible implementations

I think this should be pretty straightforward: it can go in `@linaria/atomic`'s atomizer.

PR #934, opened 2022-02-23
feat(atomic): string serialization of atoms

jpnelson:
## Motivation

Addresses https://github.com/callstack/linaria/issues/912

## Summary

Should be fairly straightforward after reading the reason why we might want this. I've also updated the documentation to reflect what's going.

## Test plan

Tested with snapshot tests and unit tests. Docs should render the same way

### DIFF
--- a/packages/atomic/src/atomize.ts
+++ b/packages/atomic/src/atomize.ts
@@ -2,6 +2,7 @@ import postcss, { Document, AtRule, Container, Rule } from 'postcss';
 import { slugify } from '@linaria/utils';
 import stylis from 'stylis';
 import { all as knownProperties } from 'known-css-properties';
+import { getPropertyPriority } from './propertyPriority';
 
 const knownPropertiesMap = knownProperties.reduce(
   (acc: { [property: string]: number }, property, i) => {
@@ -48,11 +49,13 @@ export default function atomize(cssText: string) {
     let thisParent: Document | Container | undefined = decl.parent;
     const parents: (Document | Container)[] = [];
     const atomicProperty = [decl.prop];
+    let hasAtRule = false;
 
     // Traverse the declarations parents, and collect them all.
     while (thisParent && thisParent !== stylesheet) {
       parents.unshift(thisParent);
       if (thisParent.type === 'atrule') {
+        hasAtRule = true;
         // @media queries, @supports etc.
         atomicProperty.push(
           (thisParent as AtRule).name,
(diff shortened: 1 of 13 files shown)


jpnelson:
Hi @Anber , thanks so much for looking at the other PR's I've put up!

This one addresses some issues we've come across when using atomic styles in Linaria – namely, it's confusing for people that `cx(a, cx(b, c)) !== cx(a, b, c)`. This should fix that by just providing strings. Another advantage is you don't (strictly) need to use cx with atomic styles if you're not merging – you could apply the class string directly.

Would love to get your thoughts! The issue linked provides more detailed rationale. Thanks again!

Anber (approved):

packages/babel/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": { "paths": {}, "rootDir": "src/" },
   "references": [
+    { "path": "../atomic" },
```
Anber, line 5: Looks like it's unneeded here.

Anber:
Hi @jpnelson!
It looks good to me but I don't have enough experience with atomic styles, so all decisions about it are up to you :)

packages/babel/tsconfig.json
jpnelson: Thanks! Removed

jpnelson: I realized this is being added because there's a dev dependency `@linaria/atomic`. We can keep it removed it but I think that's why it's being re-added.

jpnelson:
@Anber I added another commit to this PR, to add property priorities to address https://github.com/callstack/linaria/issues/902 . It depends on this PR so I added it to this one.

Do you think you'd be able to do a release when this goes in? (I think we'd need new versions of `@linaria/babel-preset` and `@linaria/atomic`. Thanks again for your reviews!

jpnelson:
@Anber sorry to ping again on this – do you think a merge / release is possible?

We're also testing and iterating on these features in our project. I'm hopeful this is the last major change, but if the frequency of releases is inconvenient, we could create a fork for our project and try and contribute back later in a larger chunk.

Thanks again for your time! 🙏

Anber:
@jpnelson sorry for the delay, I was in mountains for the last two weeks almost without all communications. I'll merge it and release in a couple of days.

jpnelson:
Haha that's very understandable, thank you! 🙏

packages/atomic/src/atomize.ts
```diff
@@ -1,6 +1,26 @@
 import postcss, { Document, AtRule, Container, Rule } from 'postcss';
 import { slugify } from '@linaria/utils';
 import stylis from 'stylis';
+import { all as knownProperties } from 'known-css-properties';
+import { getPropertyPriority } from './propertyPriority';
```
Anber, line 5: Hi @jpnelson
Looks like you didn't add this file to the PR

jpnelson: my mistake, sorry @Anber ! I've added it back in https://github.com/callstack/linaria/pull/950 . I think this was missing because it was originally a separate PR.

Sorry I missed that too – the precommit hooks seem extensive, but let me know if there's anything else I can run before merging – CI doesn't seem to start for me when I create PR's.

PR #950, opened 2022-03-21
feat(atomic): add property priorities

jpnelson:
This adds the missing `propertyPriority` file from https://github.com/callstack/linaria/pull/934

## Motivation
The file is missing from https://github.com/callstack/linaria/pull/934 . It adds an order of precedence on individual atoms, so that we cannot have interference between them. The docs elaborate on the precedence and the use case.

## Summary
This was left out because this was initially on another branch (this one) but cherry-picked to the other PR.

## Test plan
Tested locally


Anber:
I don't understand why CI doesn't like your commits :)

PR #951, opened 2022-03-21
fix(shaker): fix some edge cases related to specific export patterns

jpnelson:
I've tried adopting the shaker method (over extraction) and ran into some edge cases – I've added tests that explain what those edge cases look like.

## Motivation
There are three cases that weren't shaken correctly by the shaker algorithm:
1. Assigning to an export (eg. `var x = module.exports = {...}`)
2. Exporting objects (with computed keys)
3. Exporting objects directly (eg. `module.exports = {...}`) and using this object as a default import elsewhere

## Summary
This fixes it by changing the shaker algorithm to support these cases.

I think so long as we can get these new tests to pass, I'd be happy to take suggestions on exactly how we can implement the fixes. In particular, the biggest change is to support assigning to an export, for example:

```
var x = module.exports = {...}
```

The issue there is that the variable declaration `var x = ...` is marked as not-alive, and so the whole expression is shaken. This fixes it by checking variable declarations, but we could also change the `shake` part of the code to check for alive children. I think that that may have broader implications though.

Very impressed with the shaker algorithm overall, thanks for writing it!

## Test plan
Tested with snapshot tests. I believe for CI to pass, https://github.com/callstack/linaria/pull/950

### DIFF
--- a/packages/shaker/src/langs/core.ts
+++ b/packages/shaker/src/langs/core.ts
@@ -736,17 +736,19 @@ export const visitors: Visitors = {
   /*
    * SequenceExpression
    * It is a special case of expression in which the value of the whole
-   * expression depends only on the last subexpression in the list.
-   * The rest of the subexpressions can be omitted if they don't have dependent nodes.
+   * evaluates to the last subexpression in the list.
+   * In theory, some could be emitted, but we set all the subexpressions
+   * to have dependencies since there could be references in the list.
    *
    * Example:
    * const a = (1, 2, b = 3, 4, b + 2); // `a` will be equal 5
    */
   SequenceExpression(this: GraphBuilderState, node: SequenceExpression) {
-    // Sequence value depends on only last expression in the list
     this.baseVisit(node, true);
     if (node.expressions.length > 0) {
-      this.graph.addEdge(node, node.expressions[node.expressions.length - 1]);
+      for (let expression of node.expressions) {
+        this.graph.addEdge(node, expression);
+      }
     }
   },
 };
--- a/packages/babel/__utils__/strategy-tester.ts
+++ b/packages/babel/__utils__/strategy-tester.ts
@@ -306,7 +306,7 @@ export function run(
     const { code, metadata } = await transpile(
       dedent`
       import { styled } from '@linaria/react';
-
+      let external = 0;
       export const Title = styled.h1\`
         color: ${'${(external, () => "blue")}'};
       \`;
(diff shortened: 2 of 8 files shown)


Anber (approved):

Anber:
Looks perfect! Thank you!

packages/shaker/src/langs/core.ts
```diff
@@ -736,17 +736,19 @@ export const visitors: Visitors = {
   /*
    * SequenceExpression
    * It is a special case of expression in which the value of the whole
-   * expression depends only on the last subexpression in the list.
-   * The rest of the subexpressions can be omitted if they don't have dependent nodes.
+   * evaluates to the last subexpression in the list.
+   * In theory, some could be emitted, but we set all the subexpressions
+   * to have dependencies since there could be references in the list.
```
Anber, line 741: Could you please provide an example when we need to keep the full list?

jpnelson: I noticed CI was failing so I added that here: https://github.com/callstack/linaria/pull/953

an example is something like:

```
var n = 0;
export default number = (n = 5, n);
```

if we took only the last item in the sequence expression, we'd be left with

```
var n = 0;
export default n;
```

which isn't equivalent.

I created https://github.com/callstack/linaria/pull/953 which adds a proper test case for this, and I noticed CI was broken too. This updates that 🙏

jpnelson:
@Anber would a release be possible after this merges, if CI is green? Thank you again for maintaining this!

packages/shaker/src/langs/core.ts
Anber: This example should work properly without changing the shaker, because the last expression is marked as valuable by default and the first is marked as a dependency of the last one, so that both are stay alive.

Anber: Ok, I see now why you changed that behaviour :)
```
	var object = (_object = {}, (0, _defineProperty2.default)(_object, computedKeyName, 'red'), (0, _defineProperty2.default)(_object, "blue", 'blue'), _object);
```

The problem here can be solved by keeping alive all members in sequence, but it isn't the best solution. I'll release the fix today.

Anber: Ok, I see now why you changed that behaviour :)
```
	var object = (_object = {}, (0, _defineProperty2.default)(_object, computedKeyName, 'red'), (0, _defineProperty2.default)(_object, "blue", 'blue'), _object);
```

The problem here can be solved by keeping alive all members in sequence, but it isn't the best solution. I'll release the fix today.

PR #953, opened 2022-03-29
fix(shaker): fix shaker tests in CI

jpnelson:
## Motivation

I noticed CI was failing, apparently due to some tests needing to be updated.

## Summary
This doesn't change the actual behavior of shaker from what's already merged, just updates the edge cases.

## Test plan
Tested locally with `yarn test` – I think permissions might be required for CI to run on the PR too?

### DIFF
--- a/packages/babel/__tests__/depsGraph.test.ts
+++ b/packages/babel/__tests__/depsGraph.test.ts
@@ -240,6 +240,12 @@ it('SequenceExpression', () => {
   });
   expect(seqDeps).toMatchObject([
     {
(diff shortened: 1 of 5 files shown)


jpnelson:
@Anber would a release be possible after this merges, if CI is green? Thank you again!

Anber:
I've made a more generic fix for that problem. I'll release the next beta in a few hours. Thank you @jpnelson!

Anber:
`3.0.0-beta.18` has just been published

PR #969, opened 2022-05-09
Remove console.log from stylelint preprocessor

lencioni:
I noticed that our stylelint runs are logging all of our styles. I've
traced this down to this console.log call.

From what I can tell, it looks like this was added by mistake in
https://github.com/callstack/linaria/pull/876

## Motivation

I want to remove noise from my CI logs

## Summary

See above

## Test plan

N/A


Anber:
Ooops. Thank you :)

Issue #957, opened 2022-04-07
Cannot read properties of undefined (reading 'expressions')


hueindahaus:
## Environment

- Linaria version:
We use scoped packages with the following versions:
"@linaria/babel-preset": "3.0.0-beta.13",
"@linaria/core": "3.0.0-beta.13",
"@linaria/react": "3.0.0-beta.13",
"@linaria/shaker": "3.0.0-beta.13",
"@linaria/webpack-loader": "3.0.0-beta.13",
- Bundler (+ version): Webpack ^5.62.2
- Node.js version: ^16.13.0
- OS: Windows

## Description

When declaring variables with the "css" literal tag inside functions, class names can't be interpolated as expected. The following example should work:
```
const App = () => {
  const outer = css``;

  const inner = css`
    background: red;
    
    .${outer}:hover & {
      background: green;
    }
  `;

  document.body.innerHTML = `
    <div class="${outer}">
      <span class="${inner}">Hello</span> world!
    </div>
  `;
};
```

but throws :
> Module build failed (from ./node_modules/@linaria/webpack-loader/lib/index.js):
TypeError: /home/projects/webpack-webpack-js-org-v3s93t/src/index.js: Cannot read properties of undefined (reading 'expressions') at NodePath._getKey

However if I switch the positions of the declarations "inner" and "outer" it suddenly starts working:
```
const App = () => {

  const inner = css`
    background: red;
    
    .${outer}:hover & {
      background: green;
    }
  `;

  const outer = css``;

  document.body.innerHTML = `
    <div class="${outer}">
      <span class="${inner}">Hello</span> world!
    </div>
  `;
};

```
The problem with this is that typescript complains that the variable "outer" is used before it is declared.

If I declare the variables directly in the file (outside the function) it works regardless of the order of which the variables are declared.

## Reproducible Demo

A reproducible demo is provided here:
https://stackblitz.com/edit/webpack-webpack-js-org-v3s93t

The reproducible steps are essentially:
- Set up a minimal project with babel, webpack 5 (with mini-css-extract-plugin) and Linaria
- Write a function (preferably a React FC)
- Write two css tag literals to create LinariaClassNames
- Interpolate the first classname in the second

Thanks in Advance!

Anber:
Since the limitations of compile-time execution, `css` cannot be called inside components. Just extract your `inner` and `outer` from the `App`-component.

PR #971, opened 2022-05-15
fix: error when css calls reference css calls inside components

GabbeV:
## Motivation

fix: https://github.com/callstack/linaria/issues/957

## Summary

The issue was caused by the referenced template getting hoisted, converting the path already added to `state.queue` into an identifier referencing the hoisted variable. The code processing the queue expected all paths to be tagged template literals and crashed when trying to access `path.get('quasi').get('expressions')`.

Avoiding hoisting templates when referenced in other templates seem to have solved the problem.

### DIFF
--- a/packages/babel/__utils__/strategy-tester.ts
+++ b/packages/babel/__utils__/strategy-tester.ts
@@ -890,6 +890,38 @@ export function run(
     expect(metadata).toMatchSnapshot();
   });
 
+  it('should process `css` calls referencing other `css` calls inside components', async () => {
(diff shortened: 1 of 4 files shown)


Anber (approved):

Anber:
Hi @GabbeV!

Thank you!
It will fix the problem. However, I don't recommend calling `css` inside components or functions. It doesn't make any sense since you cannot use passed args/props for interpolation. It also can cause some unforeseen problems.

PR #972, opened 2022-05-16
fix(atomic): fix duplication of the same css properties in the same file

jpnelson:
## Motivation

This fixes a bug I discovered where the same property is getting overriden by other css in the same file. Test case provided for the scenario that's failing

## Summary

Relatively simple fix – this just uses the CSS text itself as a key in the metadata to ensure that it's unique per rule + declaration, between declarations too.

## Test plan

Fixed with a snapshot test

### DIFF
--- a/packages/babel/src/evaluators/templateProcessor.ts
+++ b/packages/babel/src/evaluators/templateProcessor.ts
@@ -307,7 +307,7 @@ export default function getTemplateProcessor(
       }
       const atomicRules = atomize(cssText);
       atomicRules.forEach((rule) => {
-        state.rules[rule.property] = {
+        state.rules[rule.cssText] = {
           cssText: rule.cssText,
           start: path.parent?.loc?.start ?? null,
           className: className!,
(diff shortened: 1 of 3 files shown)


packages/babel/src/evaluators/templateProcessor.ts
```diff
@@ -307,7 +307,7 @@ export default function getTemplateProcessor(
       }
       const atomicRules = atomize(cssText);
       atomicRules.forEach((rule) => {
-        state.rules[rule.property] = {
+        state.rules[rule.cssText] = {
```
juanferreras, line 310: @jpnelson great catch! I spotted this issue whilst working on the styled PR (I thought it wasn't affecting `css`, my bad) – and it seemed fixed when using `rule.className` as the key (as `atomizeFn` already creates a unique key for the rule).

is the reason you prefer `cssText` (eg. `.atm_gi_yh40bf{margin:2px;}`) over `className` (eg. `atm_gi_yh40bf`) because className can be null on certain cases (`@keyframes` only, right?)

Anber:
Hi @jpnelson!
I still don't understand, why CI doesn't run checks for your PRs :)

Anber:
Btw, I'm in the process of big refactoring of shaker and during this refactoring I've replaced Lerna with Turborepo, and that's dramatically increased the speed of (pre-commit) checks, so when I finish it, you won't be needed to skip them anymore :)

PR #973, opened 2022-05-16
feat(core): allow renaming of css template literals

jpnelson:
This allows us to have atomic css and non atomic css live in the same file. It does this by supporting renames of css imports, which exist for `styled` but don't actually exist for `css` today. Eg:

```js
import { css as myCss } from '@linaria/core';

myCss`...`;
```

previously did not work, whereas now it will. It also allows renaming of the old API:

```js
import { css as myCss } from 'linaria';
```

as well as having multiple renames coexist:

```js
import { css as coreCss } from '@linaria/core';
import { css as atomicCss } from '@linaria/atomic';
import { styled as myStyled } from '@linaria/react';
```

A few other things to note regarding backwards compatibility:

1. I changed the metadata for localName to be an object of local names for each API surface rather than a string. I don't think it should be a problem to change this format
2. I changed the name of `DetectStyledImportName` to `DetectLinariaImportName` as it now detects all linaria API's. This is technically a breaking change in the API as it's renaming an export, but I think it's an ok thing to do on master while we're on v3 beta versions. It shouldn't be hard to migrate, but might be worth making a note of in the changelog.
3. This might have an impact on https://github.com/callstack/linaria/pull/966 , but it should be possible to add to this  in a straightforward way to allow renaming of any other API's from '@linaria/atomic' too.

## Test plan
Tested with snapshot tests


Anber (approved):

Anber:
I've already done it in another branch, but don't know when it will be stable enough for merging, so lets use your implementation for a while.

PR #974, opened 2022-05-30
feat(babel): new way for detecting tags imports

Anber:
## Motivation

`@linaria/atomic` introduced the new pair of tags with their processing logic. It increased the complexity of `babel-preset`.

Another point is adding an ability to process already transpilled files.

## Summary

In this PR, I changed the way how template tags are detected and resolved:

- there are no more hardcoded packages and tags names;
- if the package imports a tag that can be processed by Linaria, its package.json should have `linaria.tags` section with a list of imported tags and links to processors that encapsulates all logic for processing those tags;
```
"linaria": {
  "tags": {
    "css": "./lib/processors/css.js",
    "styled": "./lib/processors/styled.js"
  }
},
```
- in addition to `import … from …`, this PR introduces support for transpilled imports, so it's become possible to use pure tsc compiler + `@linaria/cli` for post-processing and extracting styles from compiled code.

## Test plan

A bunch of tests were added.


ntucker:
FYI: this upgrades loader-utils but doesn't migrate the usage of it based on breaking changes. See https://github.com/callstack/linaria/issues/977

Anber:
Ooops. Thank you!

PR #976, opened 2022-06-02
feat(babel): API for custom template tags

Anber:
## Motivation

`atomic` highlighted the complexity of tags' processing logic. That part hasn't been changed since the beginning of Linaria. So, the time for changes has finally come.

## Summary

This PR introduces a new way of creating tags. Each tag (both `css` and `styled`, as well as their atomic versions) now has a `processor` that implements the abstract class `BaseProcessor`. Those processors incapsulate all tag logic and provide the API for the generic template processor inside `babel-preset`.

## Test plan

A lot of new tests were added.


Anber:
@juanferreras @jpnelson 
Hi guys!

I did some bad things here:
1. `atomize` option was removed, since it isn't needed anymore. You don't need to configure anything, just import tags form `@linaria/atomic` and it should just work. But it's actually good news :)
2. The bad news, is that I had to drop `libResolver` as well since it became unclear how to resolve tag processors in case of re-exports. What are your use cases for `libResolver`? Maybe we can find a better alternative.

And the main news is that it is become possible to make custom tags without touching Linaria's insides. The API is unstable yet and TBD, but a start has been made.

juanferreras:
> @juanferreras @jpnelson Hi guys!
> 
> I did some bad things here:
> 
> 1. `atomize` option was removed, since it isn't needed anymore. You don't need to configure anything, just import tags form `@linaria/atomic` and it should just work. But it's actually good news :)
> 2. The bad news, is that I had to drop `libResolver` as well since it became unclear how to resolve tag processors in case of re-exports. What are your use cases for `libResolver`? Maybe we can find a better alternative.
> 
> And the main news is that it is become possible to make custom tags without touching Linaria's insides. The API is unstable yet and TBD, but a start has been made.

Hi @Anber ! These last refactors you've been working on seem like fantastic iterations!

I don't have anything against either of these points, I can quote @jpnelson on why I think he went for `options.atomize`
> [...] Invert the dependency so that @linaria/atomic has the atomize function, and the postcss dependency. The atomize  function can be a part of the linaria config used by the babel preset, and if you want to use atomic styles, pass in  atomize: require('@linaria/atomic').atomize in the config

See https://github.com/callstack/linaria/pull/867#discussion_r745899890 for your original comment too. If you've managed to solve this, then this is indeed actually great news to simplify the usage!

---

I haven't used `libResolver` on my end – this is the original issue from @jpnelson https://github.com/callstack/linaria/issues/851

Anber:
> If you've managed to solve this, then this is indeed actually great news to simplify the usage!

Yep, that was the aim of this refactoring.

> I haven't used libResolver on my end – this is the original issue from @jpnelson https://github.com/callstack/linaria/issues/851

Maybe it's possible to re-import from the local project by adding something like this to `package.json`:
```
  "linaria": {
    "tags": {
      "css": "@linaria/core/processors/css.js",
      "styled": "@linaria/react/processors/styled.js"
    }
  },
```

PR #991, opened 2022-06-14
fix(esbuild): fix broken source maps

### LINKED ISSUE #898: esbuild plugin produces incorrect JSX source maps
## Environment

- Linaria version:         "@linaria/esbuild": "^3.0.0-beta.15"
- Bundler (+ version):         "esbuild": "^0.14.8"
- Node.js version:
- OS:

## Description

Pure esbuild produces correct source maps, with the JSX intact in sourcesContent. But when using the linaria plugin, sourcesContent is replaced with there result of a JSX transform (so React.createElement).

## Reproducible Demo

```
// example.jsx
const Foo = () => <div>Ok</div>;
console.log(Foo);

// build.js
require('esbuild').build({
  entryPoints: ['example.jsx'],
  bundle: true,
  outfile: 'out.js',
  sourcemap: true,
}).catch(() => process.exit(1))
```

Running `node build.js` will produce `out.js.map` with:
```
{
  "version": 3,
  "sources": ["foo.jsx"],
  "sourcesContent": ["const Foo = () => <div>Ok</div>;\nconsole.log(Foo);\n"],
  "mappings": ";;AAAA,MAAM,MAAM,MAAM,oCAAC,OAAD,MAAK;AACvB,UAAQ,IAAI;",
  "names": []
}
```

Now use build.linaria.js:

```
const linaria = require("@linaria/esbuild").default;

require("esbuild")
    .build({
        entryPoints: ["foo.jsx"],
        bundle: true,
        outfile: "out.js",
        sourcemap: true,
        plugins: [
            linaria({
                // Does not matter what you use here or if you leave it out.
                sourceMap: false,
            }),
        ],
    })
    .catch(() => process.exit(1));
```

Running `node build.linaria.js` will produce `out.js.map` with:
```
{
  "version": 3,
  "sources": ["foo.jsx"],
  "sourcesContent": ["const Foo = () => /* @__PURE__ */ React.createElement(\"div\", null, \"Ok\");\nconsole.log(Foo);\n"],
  "mappings": ";;AAAA,MAAM,MAAM,MAAsB,sBAAM,cAAc,OAAO,MAAM;AACnE,UAAQ,IAAI;",
  "names": []
}
```

Note the sourcesContent is incorrect, it should have the original JSX. Not sure how to debug this, I tried flipping sourceMap when initializing the plugin but it did not work.

silviogutierrez:
Fixes #898

### DIFF
--- a/packages/esbuild/src/index.ts
+++ b/packages/esbuild/src/index.ts
@@ -70,10 +70,19 @@ export default function linaria({
           }
         }
 
-        const { code } = transformSync(rawCode, {
+        const transformed = transformSync(rawCode, {
           ...options,
+          sourcefile: args.path,
+          sourcemap: sourceMap,
           loader,
         });
+        let { code } = transformed;
+
+        if (sourceMap) {
+          const esbuildMap = Buffer.from(transformed.map).toString('base64');
+          code += `/*# sourceMappingURL=data:application/json;base64,${esbuildMap}*/`;
+        }
+
         const result = transform(code, {
           filename: args.path,
           preprocessor,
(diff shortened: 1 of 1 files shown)


packages/esbuild/src/index.ts
```diff
@@ -93,18 +102,21 @@ export default function linaria({
         const slug = slugify(cssText);
         const cssFilename = `${filename}_${slug}.linaria.css`;
 
+        let contents = `import ${JSON.stringify(cssFilename)}; ${result.code}`;
```
silviogutierrez, line 105: I noticed everything needs to be in one line. Otherwise the maps will be off by 1 line.

I also couldn't figure out what the `import` statement here achieves. Without it, CSS still worked fine. But I left it there due to https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence

Anber: I have no idea about the `import` either :)

Anber (approved):

Anber:
Thank you, @silviogutierrez!

PR #1001, opened 2022-07-12
feat(babel): async resolvers (fixes #888 and a lot of more)

### LINKED ISSUE #888: @linaria/rollup does not support `resolve.alias` option from configuration.
## Environment

- Linaria version: "^3.0.0-beta.15"
- Bundler (+ version): vite@^2.7.3"
- Node.js version: 16
- OS: macOS

## Description

I'm using vite (that wraps rollup) to bundle my react app.

In my `vite.config.js` I have the following alias that enables `src/` absolute imports accross the app.

```
  resolve: {
    alias: [
      {
        find: 'src',
        replacement: resolve(ROOT_DIR, 'src'),
      },
    ],
  },
```

Unfortunately linaria does not seem to leverage this to properly resolve these aliased absolute paths:

Tried to tweak the `linaria.config.js` without luck:

```js
const {resolve} = require('path');

module.exports = {
  evaluate: true,
  displayName: true,
  babelOptions: {
    presets: ['@babel/preset-react', '@babel/preset-typescript'],
  },
  resolveOptions: {
    alias: [
      {
        find: 'src',
        replacement: resolve(__dirname, 'src'),
      },
    ],
  },
};
```

Anber:
## Motivation

From the beginning, Linaria worked in synchronous mode. That was the reason, why we could not directly support resolvers provided by bundlers (because all of them are async).

## Summary

This PR introduces a completely new Linaria, that was rewritten from scratch in order to support async mode. Now Linaria natively supports Rollup, Vite, esbuild and webpack. You don't even need to configure `babel-plugin-module-resolver` for aliases if your bundler already supports them. Such a big change means, that it can not be released under 3.0 branch. It also means, that the first beta of 4.0 will be pretty unstable, but only during the build stage. If your project is successfully built under 4.0, everything will be fine in production.

## Test plan

A lot of new tests were written. Examples for Rollup, Vite, esbuild and webpack 4/5 were added.


Anber:
@jpnelson I'm going to bring back `libResolver` option under the new name and a slightly different logic. https://github.com/callstack/linaria/pull/1001/files#diff-58f184d7415a8c22e56e4908527cfd1c7c28ea1c5720cf4280b2d511b25f7409R80 Is it ok for you?

PR #1031, opened 2022-07-30
feat(babel): react hooks aren't needed for evaluation

Anber:
## Motivation

The size of evaluated code can be reduced by shaking out all functions that use react hooks.

## Summary

This PR modifies the preeval stage to detect and delete not only `createElement` calls but all react functions that can be used only inside components.

### DIFF
--- a/packages/babel/src/plugins/preeval.ts
+++ b/packages/babel/src/plugins/preeval.ts
@@ -6,7 +6,7 @@ import type { BabelFile, NodePath, PluginObj } from '@babel/core';
 import type { CallExpression, Identifier } from '@babel/types';
 
 import { createCustomDebug } from '@linaria/logger';
-import type { StrictOptions, IImport } from '@linaria/utils';
+import type { StrictOptions, IImport, ISideEffectImport } from '@linaria/utils';
 import {
   collectExportsAndImports,
   getFileIdx,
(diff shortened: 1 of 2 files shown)


packages/babel/src/plugins/preeval.ts
```diff
@@ -50,14 +50,33 @@ const isBrowserGlobal = (id: NodePath<Identifier>) => {
   return forbiddenGlobals.has(id.node.name) && isGlobal(id);
 };
 
-function isCreateElement(
+const unnecessaryImports = new Set([
```
ntucker, line 53: https://reactjs.org/docs/hooks-reference.html

Don't forget the new ones:

[useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle)
[useDebugValue](https://reactjs.org/docs/hooks-reference.html#usedebugvalue)
[useDeferredValue](https://reactjs.org/docs/hooks-reference.html#usedeferredvalue)
[useTransition](https://reactjs.org/docs/hooks-reference.html#usetransition)
[useId](https://reactjs.org/docs/hooks-reference.html#useid)

Anber: I think, it's better to detect all use* methods, if some new will be added later. Will fix it tomorrow.

PR #1041, opened 2022-08-04
fix(preeval): do not remove global var in string template

### LINKED ISSUE #1039: Build failed for js literal template
## Environment
- Linaria version: 4.1.2
- Bundler (+ version): react-scripts@5.0.1
- Node.js version: v14.17.6
- OS: Mac OS (M1 chip)

## Description
There are two files, `index.js` and `literal.js`, bellow is their contents
**index.js**
```js title='index.js'
import { css } from '@linaria/core'
import { literal } from './literal'

export const style = css`
  width: 1px;
`
```
**literal.js**
```js
export const literal = `${window.location.name}`
```
When build, will get error
```
Unexpected token (1:26)
> 1 | export const literal = `${}`;
```
image

## Reproducible Demo
https://github.com/anc95/linaria-template-string-demo

anc95:
## Motivation
Should fix #1039 

## Summary
it's safe to not remove the global var inside `TemplateString`, I replace it with empty string

### DIFF
--- a/packages/utils/src/scopeHelpers.ts
+++ b/packages/utils/src/scopeHelpers.ts
@@ -127,6 +127,16 @@ export function findParentForDelete(path: NodePath): NodePath | null {
     return findParentForDelete(parent);
   }
 
+  if (parent.isTemplateLiteral()) {
+    mutate(path, (p) => {
+      p.replaceWith({
+        type: 'StringLiteral',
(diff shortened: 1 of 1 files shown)


Anber (approved):

Anber:
Thank you!

PR #1049, opened 2022-08-20
docs: added -m and -t options for CLI

ArthurStam:
## Motivation

CLI has two handy arguments that are missed in the documentation: `-t` and `-m`


Anber:
Thank you!

PR #1061, opened 2022-09-08
fix(bable): avoid using different babel versions

Anber:
## Motivation

In some cases, the ".key is not a valid Plugin property" error has happened.

## Summary

The error was caused by using different babel instances (from different node_modules). To avoid that, we can explicitly pass the one babel instance through the whole call stack.

### DIFF
--- a/.changeset/chatty-ants-eat.md
+++ b/.changeset/chatty-ants-eat.md
@@ -1,0 +1,7 @@
+---
+'@linaria/babel-preset': minor
+'@linaria/shaker': minor
+'@linaria/utils': minor
+---
+
+In some cases, different parts of babel-preset could use different versions of installed @babel/core. It caused the ".key is not a valid Plugin property" error. Fixed.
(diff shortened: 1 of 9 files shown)


Howie126313:
> f7351b0

I had the same problem, has this been resolved?

Anber:
Hi @Howie126313 
I haven't seen this problem since this PR was merged. Could you please provide a repo with this error?

Howie126313:
> Hi @Howie126313 I haven't seen this problem since this PR was merged. Could you please provide a repo with this error?

https://github.com/Howie126313/taro-linaria- 
Thanks Reply~This repo can reproduce~

Howie126313:
> Hi @Howie126313 I haven't seen this problem since this PR was merged. Could you please provide a repo with this error?

After updating all dependencies, it was solved, but this problem（https://github.com/callstack/linaria/issues/1022） occurred.

xusang:
I had the same problem

Anber:
That means that my assumption about the cause of the issue was wrong. Until I find another solution, the only way to get rid of that ".key is not a valid Plugin property" is to update all babel-related packages in your project to the latest versions.

PR #1064, opened 2022-09-09
docs: linter configuration updates

ArthurStam:
## Motivation

I think it would be nice to notify users that they have to "teach" stylelint to work with css-in-js syntax.

### DIFF
--- a/docs/LINTING.md
+++ b/docs/LINTING.md
@@ -9,7 +9,7 @@ For linting styles with [stylelint](https://stylelint.io/), we provide our custo
 You need to install `stylelint` and optionally your favorite config (such as `stylelint-config-recommended`) in your project:
 
 ```bash
-yarn add --dev stylelint stylelint-config-recommended
+yarn add --dev stylelint stylelint-config-recommended @stylelint/postcss-css-in-js postcss-syntax postcss
 ```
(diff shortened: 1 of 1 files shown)


docs/LINTING.md
```diff
@@ -23,7 +23,8 @@ Here's the example `.stylelintrc` configuration file:
   "extends": [
     "stylelint-config-recommended",
     "@linaria/stylelint"
-  ]
+  ],
+  "customSyntax": "@stylelint/postcss-css-in-js"
```
kutnickclose, line 27: (not a maintainer or anything) I think there may need to for stylelint 13/postcss 7 directions and stylelint 14/postcss 8 directions. 

Personally, when I upgraded to stylelint 14 I used the `postcss-scss` customSyntax.   @linaria/stylelint uses a processor that processes a jsx file into a css-like file so linting that with the css-in-js syntax I think wouldn't work.   But, I you can use the postcss-css-in-js syntax if you don't use the @linaria/stylelint processor i.e. add `processors: []` to remove the processor that gets added via extension from @linaria/stylelint.

Anber:
It's deprecated since @kutnickclose made the new stylelint plugin.

Issue #1047, opened 2022-08-18
The function cannot be used in a string template


weyheyhey:
## Environment

Linaria version: v4+
Bundler (+ version): Webpack 5+
Node.js version: 16+

## Description

Using a function declaration in a string template results in the error "This expression cannot be used in a template because it is a function parameter.":

```tsx
function getColor() {
  return 'red';
}

const Title = styled.h1`
  color: ${getColor()};
`;
```

At the same time, the use of function expression is correctly processed:

```tsx
const getColor = () => {
  return 'red';
}

const Title = styled.h1`
  color: ${getColor()};
`;
```

## Reproducible Demo

https://codesandbox.io/s/jolly-leftpad-6gfqz1?file=/pages/index.js

paddyobrien:
Also seeing this issue, it happens when I try to upgrade `@linaria/webpack-loader` from `3.0.0-beta.16` to `4.1.3`

paddyobrien:
The issue appears to happen here: https://github.com/callstack/linaria/blob/master/packages/babel/src/utils/collectTemplateDependencies.ts#L142

`binding.kind` is set to `hoisted` and the above error message is generated

PR #1069, opened 2022-09-21
Handle already hoisted bindings

paddyobrien:
## Motivation

Fixes https://github.com/callstack/linaria/issues/1047

## Summary

This incorrectly failed with an "unsupported error" in cases where the identifier was already hoisted. 

A couple of lines down from here we return early if the `binding.scope.parent` is falsey. So I think it's safe to let this through.

## Test plan

I wasn't able to get the test suite (or bootstrap) to run with pnpm so I haven't included any tests. 

It seems like it would be fairly easy to add a snapshot test along the lines of the following to the test file [here](https://github.com/callstack/linaria/blob/master/packages/testkit/src/babel.test.ts):

```
  it('should work with already hoisted vars', async () => {
    const { code, metadata } = await transform(
      dedent`
        function color() {
          return 'red\
        }

        export const square = css\`
          color: ${'${color()}'}
      \`;
      `,
      [evaluator]
    );

    expect(code).toMatchSnapshot();
    expect(metadata).toMatchSnapshot();
  });
```

### DIFF
--- a/.changeset/stale-bees-collect.md
+++ b/.changeset/stale-bees-collect.md
@@ -1,0 +1,7 @@
+---
+'@linaria/babel-preset': patch
+'@linaria/shaker': patch
+'@linaria/utils': patch
+---
+
+Fix function usage in string literals. Fixes #1047.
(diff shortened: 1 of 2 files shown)


Anber (approved):
Hi @paddyobrien!
Thank you for this fix!

paddyobrien:
No problem @Anber do you think it will be merged soon? We are currently blocked on upgrading to the 4.x releases due to this issue. (And I suspect we might have other issues once we get this one fixed)

Anber:
@paddyobrien I'm going to release the next version in a couple of days.

Issue #1077, opened 2022-10-11
Linaria + Webpack Dev Server Missing Changes from Files Imported by Linaria Files


jneander:
## Environment

- Linaria version: 4.1.3
- Bundler (+ version): webpack@5.74.0
- Node.js version: 16.7.0
- OS: macOS 12.6 Monterey

## Description

I have Linaria files (`*.linaria.js`, or `*.css.js` in my case) that are producing styles through `css` from `@linaria/core`. The styles incorporate values imported from other files. When using `@linaria/webpack5-loader` and running the `webpack-dev-server`, changes to the Linaria files trigger a rebuild and changes are hot-reloaded onto the page. Changes to the other file (the one being imported into the Linaria file) will also trigger a rebuild. However, the result of the build remains unchanged until the Linaria file is touched to trigger an additional rebuild.

I am not sure if this is a configuration issue that I can fix or work around, or if an internal change within linaria/webpack dependencies is needed. I hope it's the former.

## Reproducible Demo

https://github.com/jneander/linaria-issue-1077

jneander:
`@linaria/webpack5-loader` does get triggered by the change, and does produce output using the change. But the css file is not rewritten with the new changes. Same effect when using `style-loader` instead.

jneander:
Fixed with #1080.

PR #1080, opened 2022-10-13
fix(webpack): pass dependencies to cached css files

jneander:
When webpack is run with `--watch` or with `webpack-dev-server`, the generated css output will not be updated when dependencies imported into Linaria JavaScript source files, causing the resulting output to be stale.

The changes to the `ICache` interfaces are backward-compatible to avoid breaking any external implementations provided to the webpack builders using the prior versions of the interfaces.

## Motivation

This addresses #1077.

## Summary

Webpack allows adding dependencies to modules. Since the `*.linaria.js` and `*.linaria.css` files need to be in sync in order to generate the correct changes in webpack's watch mode, the css loader needs to have access to that dependency list so that _it_ can tell webpack when it needs to change. The logical place for this was in the cache provider. But rather than update the cached data type and break the contract on the api, I opted to add an optional set of methods which implementers don't need to worry about unless they need to address this same issue using their own cache implementation.

These changes apply to both `@linaria/webpack4-loader` and `@linaria/webpack5-loader`.

## Test plan

The proof-of-concept for this fix came from hacking the `@linaria/webpack5-loader` installation used in another project. When the changes are applied to the installation, the desired behavior is realized. The intended behavior has been verified using a demo project for both the webpack 4 and webpack 5 versions.

To verify the behavior, follow the test plan in the example repo linked in #1077. At step 5, the behavior described will be corrected after these changes are applied.

### DIFF
--- a/.changeset/curly-yaks-turn.md
+++ b/.changeset/curly-yaks-turn.md
@@ -1,0 +1,7 @@
+---
+'@linaria/webpack4-loader': patch
+'@linaria/webpack5-loader': patch
+'@linaria/webpack-loader': patch
+---
+
+Watch dependencies from cached css files in webpack watch mode.
--- a/docs/BUNDLERS_INTEGRATION.md
+++ b/docs/BUNDLERS_INTEGRATION.md
@@ -184,9 +184,13 @@ The loader accepts the following options:
   > ```
   > interface ICache {
   >   get: (key: string) => Promise<string>;
-  >   set: (key: string, value: string) => Promise<void>
+  >   set: (key: string, value: string) => Promise<void>;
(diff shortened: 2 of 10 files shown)


jneander:
@Anber Of course! Apologies for missing that step. 😞 I've added that with the latest changes. Do let me know if there's anything you'd like changed.

Good news for the webpack 4 version: The types in DefinitelyTyped have been updated! So I went ahead and reincorporated the changes into the `@linaria/webpack4-loader` package and verified the behavior in a demo project. I've updated the PR body to reflect that.

Anber (approved):

jneander:
Thank you, @Anber! 🎉

Anber:
@all-contributors please add @jneander  for code

PR #1085, opened 2022-10-17
WIP

ezlie-nguyen:

### DIFF
--- a/packages/server/src/collect.ts
+++ b/packages/server/src/collect.ts
@@ -22,12 +22,14 @@ const extractClassesFromHtml = (html: string): RegExp => {
         /\\|\^|\$|\{|\}|\[|\]|\(|\)|\.|\*|\+|\?|\|/g,
         '\\$&'
       );
-      htmlClasses.push(className);
+      if (className && className !== '') {
+        htmlClasses.push(`.${className}`);
+      }
     });
     match = regex.exec(html);
   }
 
-  return new RegExp(htmlClasses.join('|'), 'gm');
+  return new RegExp(`(?:[\\s]|^)(${htmlClasses.join('|')})(?=[\\s]|$)`, 'gm');
 };
 
 export default function collect(html: string, css: string): CollectResult {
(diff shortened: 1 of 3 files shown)


packages/server/src/collect.ts
```diff
@@ -22,12 +22,14 @@ const extractClassesFromHtml = (html: string): RegExp => {
...
-      htmlClasses.push(className);
+      if (className && className !== '') {
+        htmlClasses.push(`.${className}`);
+      }
     });
     match = regex.exec(html);
   }
 
-  return new RegExp(htmlClasses.join('|'), 'gm');
+  return new RegExp(`(?:[\\s]|^)(${htmlClasses.join('|')})(?=[\\s]|$)`, 'gm');
```
ezlie-nguyen, line 32: uses noncapturing groups with positive behind and positive lookahead to match the classname exactly, e.g. to have `.blah` not match `.b`

source: https://stackoverflow.com/questions/21448139/match-list-of-words-without-the-list-of-chars-around

Anber: Hi @ezlie-nguyen!

I think `[\\s]|^` and `[\\s]|$` could be replaced with `\\b`, which means the beginning or end of a word. In such case, you don't even need lookahead and groups:
```
return new RegExp(`\\b(${htmlClasses.join('|')})\\b`, 'gm')
```

ezlie-nguyen: Thank you @Anber! I can definitely make that change. I am wondering what you think about this idea vs. https://github.com/ezlie-nguyen/linaria/commit/fb727fd44cdd7622fde38140ec9528281c804b6d.

We have a specific use case where most of our CSS selectors are something like `.ltr.otherClassname` or `.rtl.otherClassname` and a wrapper component in our app has either the `ltr` or `rtl` classname. Because of the current regex, the HTML matches all of our selectors, and we don't see a performance gain from critical CSS.

Changing the regex would work but there are many edge cases to consider like pseudo-classes and combinators.

Anber: Since it doesn't break anything, I don't mind :)

ezlie-nguyen: I opened a new PR with the change here! https://github.com/callstack/linaria/pull/1102

PR #1096, opened 2022-10-23
feat(solid): initial draft implementation

raveclassic:
## Motivation

This PR introduces an initial implementation of linaria processor for solid.

## Summary

The idea is to completely eliminate any extra runtime components and transpile the code into intermediate representation suitable for further transpilation with solid (this means, the processor doesn't transpile the code down to solid's output, it should be the responsibility of the end developer to pipe transformations correctly: first goes @linaria/solid, then the code must be transformed by solid with either babel plugin, vite plugin etc.)

## Test plan
The tests are not working at the moment as it turns out it's quite hard to write any meaningful tests without the actual transpilation test. Testing snapshots seems too fragile and unmaintainable.

Instead I tried to take an approach on transpiling the code of tests with linaria+solid in a custom jest transform. It works but I'd still like to avoid testing stringified (or jsonified) output. Instead, it would be really cool to make use of JSDom to _run_ the output, render generated components and test _interaction_ with them. E.g. a reactive prop on a component must update components runtime structure etc.

At the moment I'm able to transpile the code with both jest transform (seamlessly transpiles the whole test and we can use regular TSX to write the code - huge win) and with a custom transpile step (takes a stringified input and transpiles down to solid's output - not so good as we lose TSX and still need to eval the code somehow).

Another thing to consider is that jest transform throws away generated css and I'd like to render it with JSDom so that interactions with components can be fully tested (e.g. when we set a prop on a component, component's color turns blue).
But this would require adding a _build_ step with esbuild/vite/etc and somehow integrate it into jest+jsdom - the first thing that comes to mind is a custom jest environment based on jest-environment-jsdom, this _might_ work. Also we could take a look at `@testing-library/react` and see what they are doing and try to implement the same things in the jest environment.

All these turned out to be much more work, so I stopped on that as unfortunately I don't have enough time at the moment.

Still, I'm opening the PR to demonstrate possible implementation of processor, as @Anber suggested in DM 👌

### DIFF
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -22,6 +22,7 @@
     "@types/dedent": "^0.7.0",
     "@types/node": "^17.0.39",
     "babel-preset-jest": "^28.1.0",
+    "babel-preset-solid": "^1.5.4",
     "dedent": "^0.7.0",
     "solid-js": "^1.4.8"
   },
@@ -45,7 +46,8 @@
     "esm/",
     "lib/",
     "processors/",
-    "types/"
+    "types/",
+    "linaria.config.js"
   ],
   "homepage": "https://github.com/callstack/linaria#readme",
   "keywords": [
@@ -56,11 +58,6 @@
     "styled-components"
   ],
   "license": "MIT",
-  "linaria": {
-    "tags": {
-      "styled": "./lib/processors/styled.js"
-    }
-  },
   "main": "lib/index.js",
   "module": "esm/index.js",
   "peerDependencies": {
@@ -76,11 +73,10 @@
     "build:declarations": "tsc -p tsconfig.lib.json --emitDeclarationOnly --noEmit false --outDir types",
     "build:esm": "babel src --out-dir esm --extensions '.js,.jsx,.ts,.tsx' --source-maps --delete-dir-on-start",
     "build:lib": "cross-env NODE_ENV=legacy babel src --out-dir lib --extensions '.js,.jsx,.ts,.tsx' --source-maps --delete-dir-on-start",
-    "test": "jest --config ../../jest.config.js --rootDir .",
+    "test": "jest --config ./jest.config.js --rootDir . --verbose",
     "test:dts": "dtslint --localTs ../../node_modules/typescript/lib __dtslint__",
     "typecheck": "tsc -b tsconfig.spec.json",
-    "watch": "npm run build --watch",
-    "t": "tsc --listFiles -p tsconfig.lib.json"
+    "watch": "npm run build --watch"
   },
   "sideEffects": false,
   "types": "types/index.d.ts",
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -25,6 +25,7 @@
     "babel-preset-jest": "^28.1.0",
     "babel-preset-solid": "^1.5.4",
     "dedent": "^0.7.0",
+    "esbuild": "^0.15.7",
     "jest-environment-jsdom": "^28.1.0",
     "jsdom": "^20.0.0",
     "solid-js": "^1.4.8"
@@ -80,7 +81,7 @@
     "test:dts": "dtslint --localTs ../../node_modules/typescript/lib __dtslint__",
     "typecheck": "tsc -b tsconfig.spec.json",
     "watch": "npm run build --watch",
-    "build:tests": "babel ./__tests__/processors/styled.test.tsx -o ./tmp/out.js --verbose"
+    "t": "node __tests__/processors/test.js"
   },
   "sideEffects": false,
   "types": "types/index.d.ts",
--- a/packages/solid/package.json
+++ b/packages/solid/package.json
@@ -1,0 +1,89 @@
+{
+  "name": "@linaria/solid",
+  "description": "Blazing fast zero-runtime CSS in JS library for SolidJS",
+  "version": "4.0.0",
+  "bugs": "https://github.com/callstack/linaria/issues",
+  "contributors": [
+    {
+      "name": "Kirill Agalakov",
+      "email": "raveclassic@gmail.com",
+      "url": "https://github.com/raveclassic"
+    }
+  ],
+  "dependencies": {
+    "@linaria/core": "workspace:^",
+    "@linaria/tags": "workspace:^"
+  },
+  "devDependencies": {
+    "@babel/types": "^7.18.9",
+    "@types/babel__core": "^7.1.19",
+    "@types/node": "^17.0.39",
+    "solid-js": "^1.4.8"
+  },
+  "engines": {
+    "node": "^12.16.0 || >=13.7.0"
+  },
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./types/index.d.ts",
+      "import": "./esm/index.js",
+      "default": "./lib/index.js"
+    },
+    "./*": {
+      "types": "./types/*.d.ts",
+      "import": "./esm/*.js",
+      "default": "./lib/*.js"
+    }
+  },
+  "files": [
+    "esm/",
+    "lib/",
+    "processors/",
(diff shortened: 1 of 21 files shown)


usmanyunusov:
Снимок экрана 2022-10-23 в 14 45 18

Our implementation

packages/solid/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  styled,
+  type Styled,
+  type StyledTag,
+  type StyledComponent,
+} from './styled';
+export { default as SolidStyledProcessor } from './processors/styled';
```
Anber, line 7: index must not re-export processors. Even though `styled` in Solid will not be evaluated in runtime, mixing it together with babel-related code isn't a good idea.

packages/solid/package.json
```diff
@@ -0,0 +1,92 @@
...
+  "homepage": "https://github.com/callstack/linaria#readme",
+  "keywords": [
+    "css",
+    "css-in-js",
+    "linaria",
+    "solid",
+    "styled-components"
+  ],
+  "license": "MIT",
+  "main": "lib/index.js",
```
Anber, line 62: Packages with custom tags should have `linaria.tag` section in their `package.json` with the list of exported tags and related processors.
```
"linaria": {
    "tags": {
      "styled": "./dist/processors/styled.js"
    }
  },
```

Anber (changes requested):

usmanyunusov:
Work in progress on this thread?

packages/solid/src/styled.ts
```diff
@@ -0,0 +1,40 @@
...
+  (component: 'The target component must have a className prop'): never;
+}
+
+type StyledJSXIntrinsics = {
+  readonly [TagName in keyof JSX.IntrinsicElements]: StyledTag<
+    JSX.IntrinsicElements[TagName]
+  >;
+};
+
+export declare const styled: Styled & StyledJSXIntrinsics;
```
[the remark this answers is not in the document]
raveclassic: I don't remember exactly, but there were some rare cases where typings just didn't work (props just didn't type check). I'll try to find that case.

packages/solid/src/index.ts
raveclassic: Hm, I think I just added this re-export similar to some other package, maybe `atomic`, I'll check that

packages/solid/package.json
raveclassic: This is required to run linaria directly in tests as tags in `package.json` are not detected. Maybe it's a bug, but this is the only way I could make it work in tests.

Note, since due to limitations of jsdom/cssstyle we can't write any meaningful integration test, I think it's fine to just drop them at all.

raveclassic:
@usmanyunusov yeah, in progress but I have very little time now to work on it.

raveclassic:
@Anber smth weird is happening to test snapshots in `interop` package, I have newlines removed in my output but they are expected in snapshots. This is weird since the repo is using pnpm and there should be no package version collisions. 
image

raveclassic:
I don't know why, but adding `"@babel/core": "7.18.9"` (which is the same exact version as in the repo root) to `devDependencies` in `/packages/solid/package.json` updates babel's transitive dependencies 🤷‍♂️
image

raveclassic:
@Anber https://github.com/pnpm/pnpm/issues/4868

raveclassic:
Ok I was finally able to fix this.
Seems running `pnpm i` with manually added dependencies in `package.json` is quite different from running `pnpm add -D` in the package directory. So, `pnpm add -D @babel/types@7.18.9 solid-js@^1.6.2` fixes the issue (note there're no `~`/`^` on `@babel/types` dependency). But this breaks `syncpack`, so I had to tweak `pnpm-lockfile.yml` manually and add `^` there and in `package.json`.

raveclassic:
I also dropped all tests in favor of adding a real solid example in `/examples` directory. This way we could combine tests with examples and not mess a lot with jsdom.

raveclassic:
ok, I finished the basic example built with `solid-js`, `vite`, `@linaria/vite`, and `vite-plugin-solid`, it's available at `/examples/solid`, the command to run is the regular `pnpm start`

packages/solid/package.json
raveclassic: fixed

packages/solid/src/index.ts
raveclassic: fixed

packages/solid/src/styled.ts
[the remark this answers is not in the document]
raveclassic: Here's a simple comparison of the old signature and the new one (simplified).
On the left is original, on the right - new.
image
- T1: `foo` prop is allowed in original whereas it's not allowed in the new because it's not declared;
- T2: `bar` is now not allowed in original, maybe because the component now has props; it's still not allowed in new because it's not declared;
- T3: mostly the same as `T1` - `foo` is not declared but is allowed in original; **note** - `bar` is allowed in `new` because type checker fails fast on the `foo` prop, it still rejects `bar` if `foo` is removed because none of them are declared
- T4: this case is not supported at all in original
- T5: this case is not supported at all in original
- T6: this case is not supported at all in original

@Anber how about shipping simplified typings first and see how it goes/gather some feedback later?

raveclassic:
@Anber ready for review

raveclassic:
@Anber thanks for enabling the CI on this PR 🙏 
there's something weird, I'm also on Node v16.16.0, but CI says it can't install node typings https://github.com/callstack/linaria/actions/runs/3554400920/jobs/5970646566#step:7:369 
But they are there, and locally everything is fine 🤷‍♂️

Anber (approved):

usmanyunusov:
🙏

shiro:
I just tested it quickly and don't fully understand solid, please correct me if I'm wrong.  
This seems to lose reactivity on props in some cases:
```.ts
const HelloWorldSolid = () => {
    const [count, setCount] = createSignal(0);

    return (
        <div>
            <Heading style={{"margin-top": `${count()}px`}}>
                count: {count()}
            </Heading>
            <button onClick={() => setCount((c) => c + 1)}> increment</button>
        </div>
    );
};
const Heading = styled.h1`
  background: red;
`;
```

The text in the heading updates correctly, but the margin is not applied.  
If you replace `Heading` with `h1` it works.

NightmanCZ90:
Is this PR abandoned?

ivancuric:
What's up with this PR?

ivancuric:
> just use tw/unocss + attributify much better dx: framework/language agnostic, shorter then inlining css, no need to write style merging logic for every component, just spread props.

This is just another DSL. I am looking for a way to write actual CSS.

MrFoxPro:
> > just use tw/unocss + attributify much better dx: framework/language agnostic, shorter then inlining css, no need to write style merging logic for every component, just spread props.
> 
> This is just another DSL. I am looking for a way to write actual CSS, not tailwind shorthands in made up attributes that may or may not clash with actual JSX props or HTML attributes

There are some drawbacks, but I think it's much better then using linaria. You don't need this rocket-science euristic compilier that slows down dev builds and potentially breaks your code. 

You can start with simple pure `class=""`. Or you can even try web components, so you will also co-locate your css with components. These are much better options for any projects in long-term run, because that tools are much easier to work with and maintain and language-agnostic. 
Linaria, on the other hand, has strong legacy vibes.

I spent year with unocss after linaria and it feels so much better to work with.

raveclassic:
@MrFoxPro let's keep this PR focused on _linaria_ and _solid_. This is not the right place to promote tailwind or any other framework.

raveclassic:
@Tetrax-10 @NightmanCZ90 @ivancuric I see all your comments here but unfortunately I don't have spare time to work on this. Your are kindly welcome to take on the initiative here 🙏

PR #1057, opened 2022-09-03
Fix dual packages ES module import bug

iMoses:
Fixes #904 (and possibly #1043)

Solution is based on this documentation: https://nodejs.org/api/packages.html#approach-1-use-an-es-module-wrapper

It properly configures the package entry point for both CommonJS and ESM, which is currently not the case (see #904)

Issue #1043, opened 2022-08-09
Named export 'styled' not found. The requested module '@linaria/react' is a CommonJS module


RanolP:
## Environment

- Linaria version: 
    "@linaria/react": "4.1.2",
    "@linaria/rollup": "4.1.2"
- Bundler (+ version): vite 3.0.4 with vite-plugin-ssr 0.4.18
- Node.js version: v18.7.0
- OS: Arch Linux

## Description

I'm writing a frontend with an SSR server using ESM and recently wanted to use linaria. But the linaria packages cannot be imported as an ESM module in a Node ESM environment.

## Reproducible Demo

[RanolP/node-esm-linaria-repro](https://github.com/RanolP/node-esm-linaria-repro), It does not use any bundler at all for concentrating to its root problem.

## Related Issue

- #904

Anber:
Hi @RanolP!

Probably, has been fixed in the last release. Could you please confirm it?

Anber:
Looks like we can just update the version of `@emotion/is-prop-valid`. I'll publish a fix in a couple of days.

Anber:
@RanolP released.

Anber:
I don't think it's a big deal to just fork `@emotion/is-prop-valid`. I'll do it this weekend.

PR #1097, opened 2022-10-24
fix(react): force interop check to fix @emotion/is-prop-valid esm import

andparsons:
## Motivation

I've setup a minimal codesandbox with fixed package versions that should highlight the issue: https://codesandbox.io/p/sandbox/sharp-brook-3298j7?file=%2Fpages%2Findex.tsx

Use of the `styled` function in a next.js project causes a runtime exception seen in a stack trace here https://3298j7-3000.preview.csb.app

```
TypeError: validAttr is not a function
./node_modules/@linaria/react/dist/index.mjs (18:12)
```

## Summary

Related: #1057 & #1093

It seems like a bug has been introduced after the support was added for esm exports in v4.2 where trying to import @emotion/is-prop-valid in @linaria/react

## Test plan

I've added a possible fix for this where we can pre-emptively transform the import before use in the module interop style, where we can check the typeof for the import and normalise the shape to `{default: Fn}`

### DIFF
--- a/.changeset/wild-hairs-grin.md
+++ b/.changeset/wild-hairs-grin.md
@@ -1,0 +1,5 @@
+---
+'@linaria/react': patch
+---
+
+force interop check to fix @emotion/is-prop-valid esm import
--- a/packages/react/src/styled.ts
+++ b/packages/react/src/styled.ts
@@ -65,8 +65,15 @@ function filterProps<T extends Record<string, unknown>, TKeys extends keyof T>(
     component.indexOf('-') === -1 &&
     !isCapital(component[0])
   ) {
+    /**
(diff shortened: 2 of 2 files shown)


alikazemkhanloo:
I closed that one in favor of #1043. You can open your pull request to be reviewed by maintainers. In #1043  someone wants to fork the upstream package. Not sure which way is better.

Anber:
I like that solution more than the idea of forking :)

Anber:
@all-contributors please add @andparsons for code

Anber (approved):

shivamragnar:
Is this released ?

Anber:
@shivamragnar yep

PR #1102, opened 2022-10-25
feat: add ignored and excluded classes to critical CSS extraction in `collect`

ezlie-nguyen:
## Motivation

Our codebase uses the [postcss-rtlcss](https://www.npmjs.com/package/postcss-rtlcss) package to add `.dir-ltr`, `.dir-rtl`, or `.dir` prefixes to our CSS selectors, so that most of them are in the following format:
```
.dir-ltr.otherClassname {
   padding-bottom: 12px;
}
```

Since our HTML contains the `dir-ltr` class, when we use Linaria's `collect` function, a lot of unnecessary selectors are added to the critical CSS even though `otherClassname` is not present.

Also, we have media queries in this format:
```
@media(...) {
  .dir-ltr.otherClassname {
    padding-left: 12px;
  }
  .dir-rtl.otherClassname {
    padding-right: 12px;
  }
}
```

We want to specifically exclude the RTL selector when the direction is LTR and vice versa.

## Summary

This PR proposes to add an opt-in `ignoredClasses` parameter to `collect` so that we can ignore `dir-ltr`, `dir-rtl`, and `dir` from the regex used to determine critical CSS styles when it is present in the HTML.

Also, the `excludedClasses` parameter will allow `dir-rtl` to be excluded, so that even if `otherClassname` is in the HTML, the `.dir-rtl.otherClassname` selector will not be critical.

## Test plan

Tested with unit tests for the new parameters to the `collect` function.

### DIFF
--- a/packages/server/src/collect.ts
+++ b/packages/server/src/collect.ts
@@ -83,23 +83,30 @@ export default function collect(
   };
 
   const handleAtRule = (rule: AtRule) => {
-    let addedToCritical = false;
-
-    rule.each((childRule) => {
-      if (isCritical(childRule) && !addedToCritical) {
-        critical.append(rule.clone());
-        addedToCritical = true;
-      }
-    });
-
     if (rule.name === 'keyframes') {
       return;
     }
 
-    if (addedToCritical) {
-      rule.remove();
-    } else {
-      other.append(rule);
+    const criticalRule = rule.clone();
+    const otherRule = rule.clone();
+
+    let removedNodesFromOther = 0;
+    criticalRule.each((childRule: ChildNode, index: number) => {
+      if (isCritical(childRule)) {
+        otherRule.nodes[index - removedNodesFromOther]?.remove();
+        removedNodesFromOther += 1;
+      } else {
+        childRule.remove();
+      }
+    });
+
+    rule.remove();
+
+    if (criticalRule.nodes.length > 0) {
+      critical.append(criticalRule);
+    }
+    if (otherRule.nodes.length > 0) {
+      other.append(otherRule);
     }
   };
 
--- a/packages/server/src/collect.ts
+++ b/packages/server/src/collect.ts
@@ -1,7 +1,3 @@
-/**
- * This utility extracts critical CSS from given HTML and CSS file to be used in SSR environments
- */
-
 import type { AtRule, ChildNode } from 'postcss';
 import postcss from 'postcss';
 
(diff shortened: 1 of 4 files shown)


Anber (changes requested):
It looks good to me.
Could you please run `pnpm changeset` and describe your changes?

ezlie-nguyen:
> It looks good to me. Could you please run `pnpm changeset` and describe your changes?

@Anber added! Thanks for the review 🙂

packages/server/src/collect.ts
```diff
@@ -22,48 +25,68 @@ const extractClassesFromHtml = (html: string): RegExp => {
...
-        addedToCritical = true;
+    const criticalRule = rule.clone();
+    const otherRule = rule.clone();
+
+    let removedNodesFromOther = 0;
+    criticalRule.each((childRule: ChildNode, index: number) => {
+      if (isCritical(childRule)) {
+        otherRule.nodes[index - removedNodesFromOther]?.remove();
+        // eslint-disable-next-line no-plusplus
+        removedNodesFromOther++;
```
lencioni, line 77: I think you can avoid disabling eslint here like this:
```suggestion
        removedNodesFromOther += 1;
```

packages/server/src/collect.ts
```diff
@@ -22,48 +25,68 @@ const extractClassesFromHtml = (html: string): RegExp => {
...
+  ignoredClasses: string[] = [],
+  excludedClasses: string[] = []
+): CollectResult {
   const animations = new Set();
   const other = postcss.root();
   const critical = postcss.root();
   const stylesheet = postcss.parse(css);
-  const htmlClassesRegExp = extractClassesFromHtml(html);
+  const htmlClassesRegExp = extractClassesFromHtml(html, ignoredClasses);
+  const excludedClassesRegExp = new RegExp(excludedClasses.join('|'), 'gm');
```
lencioni, line 49: Probably a good idea to escape the excludedClasses array for special regex characters.

packages/server/src/collect.ts
```diff
@@ -22,48 +25,68 @@ const extractClassesFromHtml = (html: string): RegExp => {
         /\\|\^|\$|\{|\}|\[|\]|\(|\)|\.|\*|\+|\?|\|/g,
         '\\$&'
       );
-      htmlClasses.push(className);
+      if (className !== '' && !ignoredClasses.includes(className)) {
```
lencioni, line 28: Since this is run in a couple of loops, it might not be a bad idea to convert `ignoredClasses` to a Set outside of the loops for performance.

ezlie-nguyen:
@lencioni I think I addressed your comments, mind taking another look?

packages/server/src/collect.ts
```diff
@@ -10,10 +10,18 @@ type CollectResult = {
   other: string;
 };
 
-const extractClassesFromHtml = (html: string): RegExp => {
+function escapeRegex(string: string) {
+  return string.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&');
+}
```
lencioni, line 15: I'm curious about the regex here. Did you write this by hand or pull it from somewhere?

It might be worth borrowing lodash's implementation. https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L14259-L14279

Here's their regex, which looks a little bit different than the one in this PR: https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L152

Their code comment points to https://262.ecma-international.org/7.0/#sec-patterns as the data source for this list of characters (though that inline link didn't work for me maybe it is supposed to be this one? https://262.ecma-international.org/7.0/#sec-regular-expressions-patterns)

packages/server/src/collect.ts
```diff
@@ -22,48 +30,71 @@ const extractClassesFromHtml = (html: string): RegExp => {
...
 
   return new RegExp(htmlClasses.join('|'), 'gm');
 };
 
-export default function collect(html: string, css: string): CollectResult {
+export default function collect(
+  html: string,
+  css: string,
+  ignoredClasses: string[] = [],
+  excludedClasses: string[] = []
```
jpnelson, line 47: For naming, to me, `ignored` and `excluded` are synonyms, so I have a hard time remembering which one is doing which. Is there some other name I'm not thinking of that would help clarify what these do?

jpnelson (approved):

packages/server/src/collect.ts
jpnelson: Would you consider also not using regex to do the string matching, and just doing string searching? (ie, `indexOf`)

packages/server/src/collect.ts
ezlie-nguyen: Yeah I agree the naming for this is tricky. What do you think about `ignoredClasses` and `blockedClasses`?

packages/server/src/collect.ts
ezlie-nguyen: I'm not quite following what the string searching would look like? Won't we still have to call `string.replace` to sanitize it?

re: the regex, I pulled it from somewhere, but happy to replace it with the one from lodash.

Anber (approved):

packages/server/src/collect.ts
Anber: What's about this thread? Are you going to change names or I can merge this PR?

ezlie-nguyen: Sorry for the delay, I changed the param names and this should be good to go now!

Issue #959, opened 2022-04-09
Cannot parse typescript files on transformation (Vite)


Malien:
## Environment

Project created via vite cli using react-ts preset. 

### vite.config.ts
```ts
import { defineConfig } from 'vite'
import linaria from '@linaria/rollup'
import react from '@vitejs/plugin-react'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [react(), linaria()]
})
```

- Linaria version: 3.0.0-beta.18
- Bundler (+ version): Vite ^2.9.0
- Node.js version: 16.10.0
- OS: macOS 10.15.7 (Catalina)

## Description

I suspect an issue with `shaker` that prevents it from parsing and evaluating typescript files (at least in vite environment)

I have set up a simple reproduction demo with two files. It is not exactly minimal, but representable of typical use-case.
### main.tsx
```tsx
import React from "react"
import ReactDOM from "react-dom"
import { styled } from "@linaria/react"

import { hsl, HSLColor, lighter, toCssColor } from "./color"

const baseColor = (props: { baseColor: HSLColor }) =>
  toCssColor(lighter(props.baseColor, 0.5))

const Container = styled.div<{ baseColor: HSLColor }>`
  background: ${baseColor};
`

ReactDOM.render(
  <React.StrictMode>
    <Container baseColor={hsl(120, 100, 50)} />
  </React.StrictMode>,
  document.getElementById("root")
)
```

### color.ts
```ts
export type HSLColor = {
  h: number
  s: number
  l: number
}

export function hsl(h: number, s: number, l: number): HSLColor {
  return { h, s, l }
}

export function lighter({ h, s, l }: HSLColor, additional: number): HSLColor {
  return { h, s, l: Math.min(l * (1 + additional), 100) }
}

export function toCssColor({ h, s, l }: HSLColor): string {
  return `hsl(${h}, ${s}%, ${l}%)`
}
```

This exact configuration causes this error to be thrown and logged:
```
5:21:21 PM [vite] Internal server error: /Users/yaroslav/Developer/JS/linaria-bug/src/main.tsx: An unexpected runtime error occurred during dependencies evaluation:
/Users/yaroslav/Developer/JS/linaria-bug/node_modules/@babel/core/lib/parser/index.js:93
    throw err;
    ^

SyntaxError: /Users/yaroslav/Developer/JS/linaria-bug/src/color.ts: Support for the experimental syntax 'flow' isn't currently enabled (1:8):

[...]

PR #1106, opened 2022-11-02
docs: add vite configuration

Platane:
## Motivation

It seems like users struggle to find documentation on how to configure vite + typescript + linear #959

## Summary

I added a simple vite configuration, and one that uses typescript.

I noted that `enforce: 'pre'` is added to the plugin in the example. I am not sure why so I omitted it in the examples. 

## Alternative

It could be useful to also add another entry in the [examples](https://github.com/callstack/linaria/tree/master/examples) directory, for vite + typescript + linear.

I don't think it will be interesting for the example to build the website ( as the website does not uses typescript ). Maybe include the source for a simple hello world page instead ?


Anber:
@all-contributors add @Platane for doc

Issue #1114, opened 2022-11-16
@linaria/shaker drops re-exported variables


kutnickclose:
## Environment

- Linaria version: v4.x
- Bundler (+ version): webpack
- Node.js version: 16.10.0
- OS:

## Description
In v3, this syntax was fine. In v4, it seems like the @liaria/shaker is dropping re-exported vars. 

Here is a test I wrote in babel.test.ts that is currently breaking but should work
```
  it('should not drop exported vars of renamed imports', async () => {
    const { code, metadata } = await transform(
      dedent`
      import { css } from "@linaria/core";
      import { foo3 } from "./__fixtures__/reexports";

      export const bar3 = foo3; //  <<<<<<<< this is the issue

      export const square = css\`
        color: ${'${bar3("thing")}'};
      \`;
    `,
      [evaluator]
    );

    console.log(code);
    expect(code).toMatchSnapshot();
    expect(metadata).toMatchSnapshot();
  });
```

When you run this test you get an error:
![image](https://user-images.githubusercontent.com/6117662/202049849-e3604ce2-2e2d-4321-b67c-e0fe0bf7a5e1.png)

note I also added `export const foo3 = () => 'foo3';` to `packages/testkit/src/__fixtures__/foo.js`

## Reproducible Demo

test above

PR #1116, opened 2022-11-17
Fix shaker dropping exports

kutnickclose:
## Motivation

https://github.com/callstack/linaria/issues/1114

## Summary

While upgrading from v3.x to v4.x shaker started dropping some of the renamed imports that are exported. This PR fixes that issue.

## Test plan

Added new test


packages/testkit/src/__snapshots__/babel.test.ts.snap
```diff
@@ -1819,6 +1819,24 @@ Dependencies: ./__fixtures__/complex-component
 
 `;
 
+exports[`strategy shaker should not drop exported vars of renamed imports 1`] = `
```
kutnickclose, line 1822: Should I not commit this?  Not sure how to fix what's going on with the breaking test.

Anber: In your environment it generates different class names?

Anber (approved):

PR #1156, opened 2022-12-09
Make it possible to configure ESBuild filter

blixt:
## Motivation

Currently the `@linaria/esbuild` plugins transforms all .js / .jsx / .ts / .tsx files that ESBuild can find. But for performance and pure functional reasons, it might be necessary to reduce that set of files to a bare minimum pattern. For example, at Framer we strictly only want to transform files that end with `….styles.ts`. We had our own custom ESBuild plugin, but in an attempt to switch to this plugin I ran into this as a blocker, and created this PR.

## Summary

The plugin can now be configured like this, to only transform a subset of the source files:

```typescript
build({
    // …
    plugins: [linaria({ filter: /\.styles\.ts$/ })],
    // …
})
```

## Test plan

To verify this works, just use the example config above and verify that only styles within the matching files (e.g. `Something.styles.ts`) get transformed.

### DIFF
--- a/packages/esbuild/src/index.ts
+++ b/packages/esbuild/src/index.ts
@@ -22,6 +22,7 @@ type EsbuildPluginOptions = {
 const nodeModulesRegex = /^(?:.*[\\/])?node_modules(?:[\\/].*)?$/;
 
 export default function linaria({
+  filter = /\.(js|jsx|ts|tsx)$/,
   sourceMap,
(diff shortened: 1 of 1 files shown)


blixt:
In the end this change was not useful to me, plus there's not been any activity here, so I'll close this PR as it's probably not worth merging it.

PR #1159, opened 2022-12-10
fix(vite): enforce: 'post'

MrFoxPro:
Vite should replace import aliases so `resolve` functions in bundlers should be able to correctly resolve imports.


MrFoxPro:
btw I'm not sure if this is good Idea... Maybe let user decide?

Issue #1072, opened 2022-09-28
Using `styled`, undefined values in interpolations are turned into strings


jpnelson:
## Environment

- Linaria version: 3.0.0-beta.13
- Bundler (+ version): next / codesandbox

## Description

It seems as though setting an undefined null value for css variables is at runtime turns that value into a string. For example,

```
const Heading = styled.h1`
  color: ${(props) => (props.color ? props.color : undefined)};
`;
const App = () => (
  <>
    <Heading color="blue">Hello world blue</Heading>
    <Heading>Hello world – no props</Heading>
  </>
);
```

At runtime, the first heading will be blue, and the second will be the default text color (expected). However, the heading gets `--var: undefined` set on its inline styles:

<img width="591" alt="Screen Shot 2022-09-28 at 11 40 30 AM" src="https://user-images.githubusercontent.com/1518604/192862736-581d1d57-dc44-4ed5-b4dd-1e0f93ddd45d.png">

I think this would be because we stringify the values to add units:

https://github.com/callstack/linaria/blob/master/packages/react/src/styled.ts#L164

It would seem to be a pretty simple fix to add a special case for `undefined` values at least (probably `null` too?)

I would think this is probably a bug fix and not a breaking change since I can't see anyone usefully relying on the fact that `undefined` values get stringified? But I'm interested in your thoughts too.

Thanks!

## Reproducible Demo
Demo here: https://codesandbox.io/s/angry-lovelace-3onv7p

Note that on the second header, with no color, we still get the `--variable: undefined` on the inline styles

juanferreras:
I've seen this and I agree with the change as a bug fix – ultimately is just bloating the HTML markup for no behavioral change (eg. `undefined`/`null`/`undefinedpx`/`nullpx`/etc. are all invalid at computed-value time which AFAIK equals the same output that simply skipping the inline custom property declaration)

There's already a [warnIfInvalid()](https://github.com/callstack/linaria/blob/master/packages/react/src/styled.ts#L162) method which could filter them out but is disabled in `NODE_ENV=production` (I'm guessing for perf purposes but I don't think there could be much of a difference?) 

I didn't submit a PR with the change yet because I was exploring if there could be any options for unlocking conditionality (or any way to bail out of a declaration) through a similar API, using CSS custom properties fallback values or something like the ["space toggle hack"](https://lea.verou.me/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/). 

[...]

PR #1163, opened 2022-12-16
feat: option to remove `var()` wrapper around css variables

aspirisen:
## Motivation
Linaria generates `var()` with one argument - the variable itself. However, css variables can accept default value:
``` 
var( <custom-property-name> , <declaration-value>? ) 
```
Since there is no way to set default value for linaria's variable we limit functionality of css variables.

Using the variable names without `var()` wrapper will allow you to use it as-is. For example, it is useful for theming: (this example will work when `variableNameConfig` is `dashes` )
```
import { styled } from "@linaria/react";

export const Theme = styled.div`
  --font-color: red;
`;

export const MyComponent = styled.div`
  color: var(${(props) => props.color ?? ""}, var(--font-color));
`;

function App() {
  return (
    <Theme>
      <MyComponent>Text with red color</MyComponent>
      <MyComponent color="blue">Text with blue color</MyComponent>
    </Theme>
  );
}
```
Here we set default variable in the root theme component `Theme`. Inside `MyComponent` we can use the linaria css variable name as-is and hence create our own `var` wrapper. So if we do not pass `color` prop css will take fallback value which is `red`. If we pass `color` props then it will be used

image

## Summary

In order to make it work i had to add new option `variableNameConfig` where you can control the variable formatting. The possible values are `var`, `dashes` and `raw`. 
- `var` is default and current behavior of linaria, each variable name is wrapped by `var`, i.e. `var(--mc195ga-0)`. 
- `dashes` will emit only valid css variable name, i.e. `--mc195ga-0`. 
- `raw` will emit only the variable name without `--`, i.e. `mc195ga-0`.  I added `raw` option just in case, i do not see a use case for that.

To make the idea of default css variable value work we have to check for `undefined`:

```
color: var(${(props) => props.color ?? ""}, var(--font-color));
```
Otherwise there will be `undefined` value for linaria style css variable and fallback will not work.
There is discussion about that https://github.com/callstack/linaria/issues/1072

This config affects whole linaria behavior, so in fact it is all or nothing, either you use `var` wrapper or use `dashes` in your application. However, probably you can configure webpack to have two linaria loaders one for `var` and another one for `dashes` using `test/include/exclude/.etc` features of your bundler. 

## Test plan

In order to make this feature you need to add the `variableNameConfig` options into webpack config.
```
{
  loader: "@linaria/webpack-loader",
  options: {
    displayName: true,
    variableNameConfig: "dashes",
  },
},
```

### DIFF
--- a/.changeset/three-bikes-lay.md
+++ b/.changeset/three-bikes-lay.md
@@ -1,0 +1,7 @@
+---
+'@linaria/tags': minor
+'@linaria/testkit': minor
+'@linaria/utils': minor
+---
+
+Add option to remove var() wrapper around css variables
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 ### OS Junk ###
 .DS_Store
 *~
+.history
 
 ### Node ###
 # Logs
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -26,6 +26,96 @@ module.exports = {
 
   Enabling this will add a display name to generated class names, e.g. `.Title_abcdef` instead of `.abcdef'. It is disabled by default to generate smaller CSS files.
 
+- `variableNameConfig: "var" | "dashes" | "raw"` (default: `var`):
+
+  Configures how the variable will be formatted in final CSS.
+
+  ### Possible values
+
+  #### `var`
+  Use full css variable structure. This is default behavior.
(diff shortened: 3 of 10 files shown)


Anber (approved):

Anber:
@all-contributors please add @aspirisen for code

PR #1179, opened 2023-01-11
fix: handle object patterns in exports

yepitschunked:
## Motivation

shaker did not handle object pattern exports like `export const { foo, bar } = baz()`

## Summary

The way that shaker associates exports seems to assume that the RHS of an export const assignment corresponds to exactly one export name. But this isn't true, as in the example above - `baz()`'s return will be destructured to multiple exported values. Due to this assumption, if one of the exports is not in `onlyExports`, the entire statement will be deleted. 

## Test plan

Added a unit test

### DIFF
--- a/packages/shaker/src/plugins/__tests__/shaker-plugin.test.ts
+++ b/packages/shaker/src/plugins/__tests__/shaker-plugin.test.ts
@@ -196,4 +196,19 @@ describe('shaker', () => {
     expect(code).toMatchSnapshot();
     expect(metadata.imports.size).toBe(0);
   });
+  it('should handle object patterns in exports', () => {
+    const { code } = keep(['Alive'])`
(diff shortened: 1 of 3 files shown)


packages/shaker/src/plugins/__tests__/shaker-plugin.test.ts
```diff
@@ -196,4 +196,19 @@ describe('shaker', () => {
     expect(code).toMatchSnapshot();
     expect(metadata.imports.size).toBe(0);
   });
+  it('should handle object patterns in exports', () => {
+    const { code } = keep(['Alive'])`
```
yepitschunked, line 200: this test throws a "no shaker metadata" error without the fix

Anber (approved):

Issue #1112, opened 2022-11-15
Cannot read properties of undefined (reading 'localeCompare')


nathanredblur:
## Environment
- Linaria version:
```
    "@linaria/babel-preset": "^4.3.0",
    "@linaria/core": "^4.2.2",
    "@linaria/react": "^4.3.0",
    "@linaria/webpack-loader": "^4.1.8",
```
- Bundler (+ version): "webpack": "^5.74.0"
- Node.js version: node 16.15.0
- OS: m1 mac OS 13.0.1

## Description

I get an error on compile a file. Not sure how this happens because it is pointing to a file with no any CSS (in fact I'm just installing and I not have any linaria code anywhere) 

```
ERROR in ./src/legacy/views/car/monthly_calendar_day_popup.js
Module build failed (from ../node_modules/@linaria/webpack-loader/lib/index.js):
TypeError: /Users/nathanredblur/getaround/getaround-web/www/src/legacy/views/car/monthly_calendar_day_popup.js: Cannot read properties of undefined (reading 'localeCompare')
```

This error refer to this line of code
https://github.com/callstack/linaria/blob/master/packages/utils/src/scopeHelpers.ts#L395
where the node exist but does not have `name`
```
a.node?.name.localeCompare(b.node?.name)
```

this is what the node have:
```
{
  type: 'StringLiteral',
  value: '',
  trailingComments: [],
  leadingComments: [],
  innerComments: []
}
```

## Reproducible Demo

not sure how to reproduce it.

nathanredblur:
I think I found how to reproduce it.
This file has an old backbone component where we generate a css like this:

```
export default CalendarPopup =  Backbone.View.extend({
   show() {
      popupLeft = 20;

      // this line is the problem
     this.$el.css({ left: `${popupLeft}px` });
   }
})
```

Why is linaria trying to compile this? and how can avoid it?

nathanredblur:
I found a workaround:
```
export default CalendarPopup =  Backbone.View.extend({
   show() {
      popupLeft = 20;

      // this is the workaround
     this.$el.css({ left: popupLeft+"px" });
   }
})
```

the problem is that this brake our prettier and eslint rules and CI don't allow passing this change.

shivamragnar:
So if the component is Server side rendered, Linaria tries to compile **backticks** template literals which is weird. Try avoiding usage of **backticks** while this issue is fixed.

My code also had the same issue 
```js
 // Linaria tries to compile this
[...]

PR #1180, opened 2023-01-11
fix: lookup referencedIdentifiers after tree mutations

yepitschunked:
## Motivation
https://github.com/callstack/linaria/issues/1112

This `localeCompare` error is very strange as identifiers should always have a name property. I did some debugging and found that the referencedIdentifiers array had paths pointing to random nodes like StringLiterals. I suspect the issue is that we look up the paths prior to mutating the AST, and the mutations end up leaving the paths pointing to invalid data. 

## Summary

Given that we don't actually use referencedIdentifiers until after the mutations are finished, I think a fix could be to just move this variable to after the mutations.

## Test plan

I don't have a good way of unit testing this change, but it did fix the localeCompare errors in our build.

### DIFF
--- a/packages/utils/src/scopeHelpers.ts
+++ b/packages/utils/src/scopeHelpers.ts
@@ -368,7 +368,6 @@ function removeWithRelated(paths: NodePath[]) {
 
   const affectedPaths = actions.map(getPathFromAction);
 
-  let referencedIdentifiers = findIdentifiers(affectedPaths, 'referenced');
   const referencesOfBinding = findIdentifiers(affectedPaths, 'binding')
     .map((i) => (i.node && getScope(i).getBinding(i.node.name)) ?? null)
     .filter(isNotNull)
@@ -390,6 +389,7 @@ function removeWithRelated(paths: NodePath[]) {
   });
 
   removeWithRelated(referencesOfBinding);
+  let referencedIdentifiers = findIdentifiers(affectedPaths, 'referenced');
 
   referencedIdentifiers.sort((a, b) =>
     a.node?.name.localeCompare(b.node?.name)


packages/utils/src/scopeHelpers.ts
```diff
@@ -368,7 +368,6 @@ function removeWithRelated(paths: NodePath[]) {
 
   const affectedPaths = actions.map(getPathFromAction);
 
-  let referencedIdentifiers = findIdentifiers(affectedPaths, 'referenced');
   const referencesOfBinding = findIdentifiers(affectedPaths, 'binding')
```
yepitschunked, line 371: we could possibly move this as well, although it did not seem to cause issues in our build like referencedIdentifiers did.

juanferreras (approved):
I have a production project where I can replicate this issue (when upgrading to v4), and this requested change does fix it. Thanks for the contribution, @yepitschunked!

@Anber curious to see your thoughts here when you get a chance – thanks!

Anber:
> Given that we don't actually use referencedIdentifiers until after the mutations are finished, I think a fix could be to just move this variable to after the mutations.

We don't use it, but we must find referenced identifiers before deleting them from the code. Otherwise, there will be nothing to find, and the shaker won't be able to detect dead code.

juanferreras:
@Anber noted – in case it's resourceful, I've isolated at least 1 case where the issue can be seen.  See https://stackblitz.com/edit/linaria-v4-localecompare?file=src%2Fapi.js

I'm not sure if this is the same one multiple folks are running into / or if there are several conditions where dead code elimination can be causing similar issues.

```
TypeError: /home/projects/linaria-v4-localecompare/src/api.js: Cannot read properties of undefined (reading 'localeCompare')
    at Array.sort (<anonymous>)
    at transformFile.next (<anonymous>)
    at run.next (<anonymous>)
    at Generator.next (<anonymous>)
```

Anber:
@juanferreras, thank you! I'll try to figure out what happened there.

Anber:
I have to take a closer look since it can be a symptom of a more complex problem, but maybe it will be enough to move `referencedIdentifiers.sort…` just after `let referencedIdentifiers…`

Anber (approved):

PR #1192, opened 2023-01-18
fix(babel): update cosmiconfig so linaria.config.cjs works

agriffis:
## Motivation

New vite projects (made with `pnpm create vite`, for example) have `"type": "module"` in `package.json`. That causes loading `linaria.config.js` to fail because of CJS/ESM mismatch.

## Summary

Upgrade cosmiconfig to 8.x. The support for `.cjs` was added in 7.0.0 but no reason to hold back from version 8.

## Test plan

I updated the code for the breaking changes in cosmiconfig 6.x, see https://github.com/davidtheclark/cosmiconfig/blob/main/CHANGELOG.md#600

I don't think we need any new tests in linaria, it's probably sufficient that the existing tests continue to pass. Anything specific for `.cjs` would just reduplicate what cosmiconfig tests upstream.

### DIFF
--- a/packages/babel/package.json
+++ b/packages/babel/package.json
@@ -44,7 +44,7 @@
     "@linaria/shaker": "workspace:^",
     "@linaria/tags": "workspace:^",
     "@linaria/utils": "workspace:^",
-    "cosmiconfig": "^5.1.0",
+    "cosmiconfig": "^8.0.0",
     "find-up": "^5.0.0",
     "source-map": "^0.7.3",
     "stylis": "^3.5.4"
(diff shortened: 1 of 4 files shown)


packages/babel/package.json
```diff
@@ -44,7 +44,7 @@
     "@linaria/shaker": "workspace:^",
     "@linaria/tags": "workspace:^",
     "@linaria/utils": "workspace:^",
-    "cosmiconfig": "^5.1.0",
+    "cosmiconfig": "^8.0.0",
```
layershifter, line 47: @agriffis please remove `"@types/cosmiconfig": "^5.0.3",` as `cosmiconfig` now provides own types, https://www.npmjs.com/package/@types/cosmiconfig/v/6.0.0

Anber (approved):

PR #1197, opened 2023-01-30
fix(babel): do not parse json as js

kcover:
## Motivation
#1182 

## Summary
Added a check to prepare-for-eval that will ignore json files. Another option would be to 
add the `.json` extension to the default regex for ignored files in [loadLinariaOptions](https://github.com/callstack/linaria/blob/master/packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts#L40) but I'm not sure if that gets overridden by custom rules and I'm not sure we'd want that.

This isn't ready for merge yet because I've encountered a new error that I was hoping I could get some help with. See the read me in this [repo](https://github.com/kcover/babel-playground-app) I made to reproduce the issue.

## Test plan
Use the repo I linked above to reproduce the issue, then use `yarn link` to utilize this branch in that project and verify that the json parsing error no longer occurs.

### DIFF
--- a/.changeset/unlucky-worms-complain.md
+++ b/.changeset/unlucky-worms-complain.md
@@ -1,0 +1,5 @@
+---
+'@linaria/babel-preset': patch
+---
+
+Avoid parsing json as js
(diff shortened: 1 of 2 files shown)


AndrewVorobev:
Any news with it? I have the exact same problem in my project

Anber:
Hi @kcover 
It looks like the problem with `displayName.charAt is not a function` has gone. Could you please confirm it?

Anber (approved):

usmanyunusov:
@Anber I'm waiting for a change

PR #1200, opened 2023-02-03
poc: use SWC in shaker for minification

layershifter:


heikir:
@layershifter any updates or future plans regarding this PR? We would love to help you get it across the finish line since we would love to adopt linaria, but no support for turbopack is currently holding us back :)

layershifter:
@heikir there are no plans to use this implementation to go forward. @Anber is currently looking into `vite-node` usage to make things faster and simplify internals.

For Turbopack support, please open an issue.

PR #1210, opened 2023-02-10
fix(rollup): don't parse external libraries

kcover:
## Motivation
Linaria fails to transform code that contains external imports when used in rollup alongside certain plugins like node-resolve. See [this bug reproduction repo](https://github.com/kcover/linaria-rollup-resolve-repro) for an example.

## Summary
The linaria rollup plugin as it exists expects any external imports to be resolved as 'null', and passes that along to the babel plugin, which knows to ignore such imports. It doesn't expect an object indicating an external dependency to be returned (See [the rollup resolveId hook docs](https://rollupjs.org/plugin-development/#resolveid) to see when this might occur). When that happens, the linaria plugin returns the `id` property of the object, which is often just the name of the import (e.g. 'react'). The babel-preset's transform function expects this to be a path to the import so it can read it in, but we don't need to do this in the case of an external dependency. This PR adds a check to handle that case where the rollup plugin receives an object indicating an external dependency.

## Test plan
I'd love to set up an automated test for this but that would be a lot of effort on my own, given how unfamiliar I am with this project. I don't see any tests already set up for the rollup plugin or I would've added some tests in there. If I can get some help/direction I'd be happy to get some automated tests started. 

In the mean time, you can pull down [the reproduction repo](https://github.com/kcover/linaria-rollup-resolve-repro) and this PR, and use yarn link to try out this change. If you have trouble using yarn link reach out to me and I can help you out.

### DIFF
--- a/packages/rollup/src/index.ts
+++ b/packages/rollup/src/index.ts
@@ -18,7 +18,7 @@ import type {
   Result,
 } from '@linaria/babel-preset';
 import { createCustomDebug } from '@linaria/logger';
-import { getFileIdx } from '@linaria/utils';
+import { getFileIdx, syncResolve } from '@linaria/utils';
 import type { Plugin as VitePlugin } from '@linaria/vite';
 import vitePlugin from '@linaria/vite';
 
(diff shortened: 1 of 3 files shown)


packages/rollup/src/index.ts
```diff
@@ -64,6 +64,13 @@ export default function linaria({
         const resolved = await this.resolve(what, importer);
         if (resolved) {
           log('resolve', "✅ '%s'@'%s -> %O\n%s", what, importer, resolved);
+
+          if (resolved.external) {
+            // ignore external modules, they don't need to be processed
+            // because external modules shouldn't contain any CSS in JS
+            return null;
```
Anber, line 71: Unfortunately, they must be processed because they can contain helpers or constants used in CSS.

Anber (approved):

Anber:
Hi @kcover. I hope this will fix your issue.

victordidenko:
Is there any workaround exists until this fix is published?

PR #1225, opened 2023-03-13
loadLinariaOptions: fix node_modules regexp for Windows

nstepien:
## Motivation

Windows compatibily.

## Summary

One regexp used forward slashes only, while filenames have blackslashes on Windows.

## Test plan

Ran locally.

### DIFF
--- a/packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
+++ b/packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
@@ -12,9 +12,11 @@ export type PluginOptions = StrictOptions & {
 const explorerSync = cosmiconfigSync('linaria');
 
 const cache = new WeakMap<Partial<PluginOptions>, StrictOptions>();
+const defaultOverrides = {};
+const nodeModulesRegExp = /[\\/]node_modules[\\/]/;
 
 export default function loadLinariaOptions(
-  overrides: Partial<PluginOptions> = {}
+  overrides: Partial<PluginOptions> = defaultOverrides
 ): StrictOptions {
   if (cache.has(overrides)) {
     return cache.get(overrides)!;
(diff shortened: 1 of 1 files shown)


packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
```diff
@@ -12,9 +12,11 @@ export type PluginOptions = StrictOptions & {
 const explorerSync = cosmiconfigSync('linaria');
 
 const cache = new WeakMap<Partial<PluginOptions>, StrictOptions>();
+const defaultOverrides = {};
+const nodeModulesRegExp = /[\\/]node_modules[\\/]/;
```
nstepien, line 16: I've extracted and made the regexp static, which should improve performance ever so slightly.

packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
```diff
@@ -12,9 +12,11 @@ export type PluginOptions = StrictOptions & {
...
 
 export default function loadLinariaOptions(
-  overrides: Partial<PluginOptions> = {}
+  overrides: Partial<PluginOptions> = defaultOverrides
```
nstepien, line 19: `= {}` would create a new object every time `overrides` is undefined, which wouldn't work as a cache key.

packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
```diff
@@ -37,13 +39,13 @@ export default function loadLinariaOptions(
...
         // The old `ignore` option is used as a default value for `ignore` rule.
-        test: ignore ?? /[\\/]node_modules[\\/]/,
+        test: ignore ?? nodeModulesRegExp,
         action: 'ignore',
       },
       {
         // Do not ignore ES-modules
         test: (filename, code) => {
-          if (!/\/node_modules\//.test(filename)) {
+          if (!nodeModulesRegExp.test(filename)) {
```
nstepien, line 48: Shouldn't the `!` be removed? Don't we want to ignore node modules?
```suggestion
          if (nodeModulesRegExp.test(filename)) {
```

packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
Anber: Nice catch 👍

packages/babel/src/transform-stages/helpers/loadLinariaOptions.ts
Anber: No, it shouldn't.  We can't just ignore node_modules because we have to process libraries which were transpiled esm, so we made a dirty check here: if a processed file in node_modules and has something that looks like an import or export statement, we will not ignore it.

Anber (approved):

PR #1233, opened 2023-04-03
fix(atomic,core,react): simplify Jest configuration

Anber:
## Motivation

Prior to this pull request, using Jest with Linaria required the inclusion of a Babel plugin to extract styles and generate class names.

## Summary

This pull request eliminates the need to include the @linaria/babel-preset when using Jest. All calls to styled and css will now be replaced with runtime versions that generate sequential class names. This simplification reduces the risk of configuration issues and streamlines the process of integrating Jest with Linaria.

### DIFF
--- a/.changeset/proud-boats-study.md
+++ b/.changeset/proud-boats-study.md
@@ -1,0 +1,7 @@
+---
+'@linaria/atomic': patch
+'@linaria/core': patch
+'@linaria/react': patch
+---
+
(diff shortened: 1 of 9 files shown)


nstepien:
You could also check for `process.env.NODE_ENV === 'test'`, that'd also make it compatible with vitest and maybe more test runners?

Anber:
@nstepien hm. It's definitely better. Thank you :)

PR #1235, opened 2023-04-15
Run Node.js CI on windows

nstepien:
## Motivation

Linaria has had paths bug on Windows in the past, so running Linaria checks on Windows in addition to Linux may help catch such bugs.

## Summary

I've tweaked `.github/workflows/check.yml` to add windows to the os matrix.

## Test plan

CI

---

CI fails right now because... of a path bug:
```
Error: re: esbuild-example:build: X [ERROR] Could not resolve "D:\nlinarialinariawebsite\nssetslinaria-logomark.svg"
. prepare: esbuild-example:build: 
. prepare: esbuild-example:build:     linaria:Hero_1v0yfuk.linaria.css:1:106:
. prepare: esbuild-example:build:       1 │ ...ound-image:url(D:\a\linaria\linaria\website\assets\linaria-logom...
. prepare: esbuild-example:build:         ╵               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
. prepare: esbuild-example:build: 
. prepare: esbuild-example:build:   You can mark the path "D:\nlinarialinariawebsite\nssetslinaria-logomark.svg" as external to exclude it from the bundle, which will remove this error.

```
I don't know if this is a bug in esbuild or linaria though.
https://github.com/callstack/linaria/blob/4b75b2de03642ecf1aacd217305fecd32e4fe1ad/website/src/components/Hero.jsx#L43


yume-chan:
I got the same issue when checking out Linaria source code on Windows. I tested some other bundlers:

* vite: `asyncResolve` returns paths with forward slashes (`url(D:/dev/linaria/...`) so no issue.
* Webpack: `asyncResolve` returns paths with forward slashes (same as esbuild), but still bundles correctly.

My guess is that esbuild expects raw (escaped) strings like they are being read from disk, but Webpack expects unescaped strings.

Anber (approved):

Issue #1226, opened 2023-03-14
Broken build when using react-redux


PierreGUI:
## Environment
```json
"@linaria/react": "^4.3.6",
"@linaria/core": "^4.2.8",
"@linaria/webpack-loader": "^4.1.15",
"webpack": "^5.76.1",
```

## Description
Using react-redux and Linaria breaks build: `EvalError: (0 , _batch.setBatch) is not a function in ...`

## Reproducible Demo
https://stackblitz.com/edit/linaria-bug-ekrflt?file=src/index.jsx

PierreGUI:
I tried a whole new setup with Vite, but the build also fails when dealing with react-redux wrapped components:
```
EvalError: (0 , _useSelector.initializeUseSelector) is not a function in
| /node_modules/react-redux/es/index.js
    at Module.evaluate (/node_modules/@linaria/babel-preset/lib/module.js:355:13)
    at require.Object.assign.ensure (/node_modules/@linaria/babel-preset/lib/module.js:295:13)
 ```

**Comment by PierreGUI** [comment]
I've tried to find earlier versions of Linaria or Babel that would compile the reproduction repo.
- I went back all the way to my current configuration: Linaria 2 (supports babel/core 7.13 max) ✅  
- Also, Linaria 3 (airbnb) builds successfully (latest babel) ✅ ([sample repo](https://stackblitz.com/edit/linaria-bug-qpalwz?file=src/index.jsx))
- Starting from Linaria 4, I run into the issue documented above (I tried different babel versions) :x:

**Comment by Anber** [comment]
Hi @PierreGUI!

This https://github.com/callstack/linaria/pull/1237 should fix it.

**Comment by PierreGUI** [comment]
I built Linaria locally and I confirm it fixes my issue. Thank you so much!

**Comment by PierreGUI** [comment]
```json
    "@babel/preset-react": "7.12.13",
    "@linaria/babel-preset": "4.4.5",
    "@linaria/core": "4.2.10",
    "@linaria/shaker": "^4.2.11",
    "@linaria/webpack-loader": "4.1.17",
```

After the latest update, this error came back:
```
ERROR in ./Parent.view.tsx
Module build failed (from ./node_modules/@linaria/webpack5-loader/lib/index.js):

EvalError: (0 , _batch.setBatch) is not a function in
| ./node_modules/react-redux/es/index.js
| ./Component.redux.ts
| ./Parent.view.tsx
[...]

PR #1237, opened 2023-04-24
fix(babel-preset,shaker): do not delete exported functions even if its body was deleted

Anber:
## Motivation

Linaria currently has a bug that leads to the deletion of exported functions from some modules when their body contains restricted code, such as JSX, React hooks, or browser API. (see #1226)

## Summary

This pull request addresses the aforementioned bug by replacing the problematic functions with an empty one instead of deleting them. By doing so, we preserve the exported functions while effectively neutralizing the restricted code within them. This approach maintains the original structure of the module exports without causing any unexpected side effects or errors.


packages/testkit/src/utils/removeWithRelated.test.ts
```diff
@@ -130,4 +132,42 @@ describe('removeWithRelated', () => {
...
+
+      export const testArrow1 = (arg) => {
+        /* remove */console.log(arg);
+      }
+
+      const testArrow2 = (arg) => (
+        /* remove */testFn = arg
+      )
+
+      export default function testDefaultFn(arg) {
```
chentsulin, line 166: Hi @Anber, is this export default function got accidentally deleted? Is this correct?

Issue #1242, opened 2023-05-04
Linaria crashes with `EvalError: Unexpected token 'export'`


nstepien:
## Environment

- Linaria version: `@linaria/webpack5-loader@4.1.16`
- Bundler (+ version): `webpack@5.82.0`
- Node.js version: 20.0.0
- OS: win11

## Description

Linaria crashes with this error with the minimal reproducible demo below:
```
> webpack

assets by status 1.85 KiB [cached] 1 asset
./main.js 39 bytes [built] [code generated] [1 error]

ERROR in ./main.js
Module build failed (from ./node_modules/@linaria/webpack5-loader/lib/index.js):
D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:355
      throw new EvalError(`${e.message} in${callstack.join('\n| ')}\n`);
      ^

EvalError: Unexpected token 'export' in
| D:\repos\vite-bug\node_modules\clsx\dist\clsx.m.js
| D:\repos\vite-bug\node_modules\react-data-grid\lib\bundle.js
| D:\repos\vite-bug\main.js

    at Module.evaluate (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:355:13)
    at require.Object.assign.ensure (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:295:13)       
    at D:\repos\vite-bug\node_modules\react-data-grid\lib\bundle.js:21:36
    at D:\repos\vite-bug\node_modules\react-data-grid\lib\bundle.js:339:3
    at Script.runInContext (node:vm:134:12)
    at Module.evaluate (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:342:14)
    at require.Object.assign.ensure (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:295:13)       
    at D:\repos\vite-bug\main.js:3:45
    at D:\repos\vite-bug\main.js:7:3
    at Script.runInContext (node:vm:134:12)
    at Module.evaluate (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\module.js:342:14)
    at evaluate (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\evaluators\index.js:20:5)
    at evalStage (D:\repos\vite-bug\node_modules\@linaria\babel-preset\lib\transform-stages\2-eval.js:26:45)
[...]

PR #1245, opened 2023-05-05
fix(babel): fixed esm detection in node_modules

Anber:
## Motivation

Some files in `node_modules` weren't detected as ESM. See #1242 

## Summary

The regular expression used to detect the keywords `export` or `import` contained an error that resulted in a false negative for ESM files that did not have any imports.

### DIFF
--- a/.changeset/blue-eagles-burn.md
+++ b/.changeset/blue-eagles-burn.md
@@ -1,0 +1,5 @@
+---
+'@linaria/babel-preset': patch
+---
+
+Better support for ES-modules in node_modules (fixes #1242)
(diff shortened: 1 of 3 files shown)


nstepien:
I wonder if this is safe, the `import` or `export` words could be in a string in the code for example.
Using babel's parser to get the AST may be safer, no?

Anber:
False positive causes just a bit more work, whereas false negative causes errors during the evaluation stage. So, false positive cases are not a big problem.
AST will be safer but more expensive cause we need to parse each file. Theoretically, we can parse only files that contain `import` or `export` and then reuse parsed AST, but it requires more complicated rules definition API.

PR #1256, opened 2023-05-30
Fix: ERROR Cannot find module 'xxx' with esbuild 0.17 and linaria 4.2

sebasjm:
Fix: ERROR Cannot find module './style.js'

## Motivation

Found this error in the logs

```
[stage-1:async-resolve] ❌ cannot resolve ./style.js in /home/sebasjm/Work/linaria/src/SimpleButton.tsx: Error: Must specify "kind" when calling "resolve"
```

With these dependencies

```
    "@linaria/core": "^4.2.10",
    "@linaria/esbuild": "^4.2.11",
    "@linaria/react": "^4.3.8",
    "esbuild": "^0.17.19",
```

## Summary

I checked the esbuild documentation and looks that there is a missing parameter. 
I patched locally and it worked for me.

## Test plan

I made a repo to reproduce and fix the problem 
https://github.com/sebasjm/linaria-esbuild-patch

### DIFF
--- a/.changeset/esbuild-linaria-cant-find-module.md
+++ b/.changeset/esbuild-linaria-cant-find-module.md
@@ -1,0 +1,5 @@
+---
+'@linaria/esbuild': patch
+---
+
+Cannot find module 'xxx' with esbuild 0.17 and linaria 4.2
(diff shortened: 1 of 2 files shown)


Anber (approved):

Anber:
@sebasjm thank you! The fix will be released later this week.

PR #1263, opened 2023-06-25
fix(vite): css `url()` resolve error

wmzy:
## Motivation

 [Relative path](https://vitejs.dev/guide/assets.html#importing-asset-as-url) in  css like: `url(../assets/img.png)` has been resolved error. This pr fixed it.

## Summary

Vite support [relative path](https://vitejs.dev/guide/assets.html#importing-asset-as-url) in  css like: `url(../assets/img.png)`.

The `cssId` was previously resolved to `cssId`. Which may cause `../assets/img.png` resolve error.

So we should resolve the `cssId` to a file which is same directory of source file.

### DIFF
--- a/.changeset/vite-url-resolve.md
+++ b/.changeset/vite-url-resolve.md
@@ -1,0 +1,5 @@
+---
+'@linaria/vite': patch
+---
+
+Fix vite css `url()` resolve error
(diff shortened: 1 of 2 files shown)


Anber:
Hi @wmzy!
Something went wrong with CI. Could you please check it?

wmzy:
Hi @Anber , problem has been resolved.

By the way, do you consider to upgrade the pnpm lockfile version to 6 so that contributors need not downgrade their pnpm version.

Anber:
Hi @wmzy!
Yeah, it's a good idea. Thank you!

PR #1292, opened 2023-07-20
Ignore babel erroneously thinking identifiers in unary expressions are bindings

GabbeV:
Babel seem to think that identifiers in unary expressions are bindings. There is already a workaround for this in the codebase but only for void.

This PR tweaks the workaround that was already there for void to handle all unary expressions.

Here is a stackblitz reproducing the issue with the + unary operator
https://stackblitz.com/edit/linaria-bug-kb4izi?file=src%2Findex.js


Anber:
It looks scary, but tests say that we are safe :)

GabbeV:
Some more context on why i think this change is correct.

The bug in babel comes from this function:

https://github.com/babel/babel/blob/fd2956fd40a550a4fabbd5df5c707ac6022ab20e/packages/babel-types/src/validators/isBinding.ts#L6

It checks if the node is in a position of its parent where bindings normally are.
To do this it uses getBindingIdentifiers.keys which erroneously lists the "arguments" property of UnaryExpression as a possible location for bindings.

https://github.com/babel/babel/blob/fd2956fd40a550a4fabbd5df5c707ac6022ab20e/packages/babel-types/src/retrievers/getBindingIdentifiers.ts#L112

As far as i know no unary operator declare bindings in JS.

The bug originates from this commit where babel for some reason needed the delete operator to act as if it were declaring a binding https://github.com/babel/babel/commit/388133642d0d573087742cb17179772a614cbdc5

Anber:
@GabbeV thank you for the explanation!

> The bug originates from this commit where babel for some reason needed the delete operator to act as if it were declaring a binding

In terms of Babel, binding is something that modifies a value, so `delete foo.bar` is the binding for `foo`, I guess.
Maybe let's add `delete` as an exception?

Anber (approved):

GabbeV:
Thank you for merging!

After looking at closer isBinding and getBindingIdentifiers.keys i find my self quite confused as to what is considered a binding by that function.

The usage in linaria in [removeWithRelated](https://github.com/callstack/linaria/blob/ae3727f9a498b6f0e50edd68a24a1fd35f1a12cb/packages/utils/src/scopeHelpers.ts#L474C15-L474C15) seem to be looking for where the variables are declared and that is what i thought binding meant in babel.

isBinding however returns true for multiple other reasons.
All these identifiers return true.
image

The most problematic is the unary math operators as they are reasonable to expect in linaria interpolations but the others will also cause linaria to misbehave even after this fix.

Replacing ``css`width: ${+size}px;`;`` in my stackblitz with any of the following will cause linaria to crash in the same way.

```
css`width: ${typeof size}px;`;
css`width: ${delete size}px;`;
css`width: ${void size}px;`;
css`width: ${++size}px;`;
css`width: ${size = 1}px;`;
css`width: ${(() => { for (size in {}) { } return 1 })()}px;`;
css`width: ${(() => { for (size of {}) { } return 1 })()}px;`;
```

My conclusion is that this PR probably solves the most urgent issues but isBinding should probably be replaced with something else as it doesn't do what the code expect.

Anber:
To make it crazier, Babel has one more conception called constant violation. So, a place where a variable is declared is called binding, places, where the value is changed, are called constant violations, and other references are just references.

Anyway, thank you for such an interesting case! I'll try to dig it out.

Anber:
Honestly, it doesn't look like a problem with unary expressions because it works without `size += 0` and `class2`. There is something more interesting.

GabbeV:
`size += 0` is only there to force linaria to evaluate it. Without it Babel considers it constant.
Using an imported variable gives the same result.

PR #1323, opened 2023-08-11
docs: Update codesandbox example

PrettyCoffee:
Your current codesandbox example seems to be out of date. Since I happened to create one for testing purposes, I thought you might like to use it. :)


Anber (approved):

Anber:
Thank you @PrettyCoffee!

PR #1325, opened 2023-08-17
fix: add support for params in dynamic imports

layershifter:
## Motivation

Follows #1318, add support for ``import(`${variable}`)``.

## Test plan

New tests added.


packages/utils/package.json
```diff
@@ -34,7 +34,6 @@
     "@babel/core": "^7.22.9",
     "@babel/generator": "^7.22.9",
     "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
-    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
```
layershifter, line 37: Was used only in `packages/utils/src/options/dynamic-import-noop.ts`, `packages/utils/src/options/dynamic-import-noop.ts` is a dead file

Anber (approved):

PR #1330, opened 2023-08-24
feat(babel): support for dynamic imports and requires

Anber:
## Motivation

```
const url = "./__fixtures__/FOO";
const foo2 = require(url.toLowerCase()).foo2;
```

## Summary

Dependencies missed during the preparation stage can now be handled during evaluation.

### DIFF
--- a/.changeset/perfect-onions-play.md
+++ b/.changeset/perfect-onions-play.md
@@ -1,0 +1,6 @@
+---
+'@linaria/babel-preset': patch
+'@linaria/testkit': patch
+---
+
+Add support for `import` and `require` calls with dynamic arguments.
--- a/packages/babel/src/index.ts
+++ b/packages/babel/src/index.ts
@@ -15,10 +15,9 @@ export { slugify } from '@linaria/utils';
 export { default as preeval } from './plugins/preeval';
 export { default as withLinariaMetadata } from './utils/withLinariaMetadata';
 export { default as Module, DefaultModuleImplementation } from './module';
-export { default as transform, withDefaultServices } from './transform';
+export { default as transform } from './transform';
 export * from './types';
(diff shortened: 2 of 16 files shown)


PierreGUI:
Tested this branch with my build, it bursts a ton of warnings:
```
[linaria] ... has a module-resolver plugin in its babelrc, but it is not presentin the babelOptions for the linaria plugin. This works for now but will be an error in the future.Please add the module-resolver plugin to the babelOptions for the linaria plugin.
```
Note: Adding the plugin to linaria's babelOptions in webpack config doesn't seem to help.

Then, the build stops and node crashes with a **`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`** 🥵

Anber:
Ok, there are at least three different problems:
1. The warning should be shown at most once.
2. It should not be shown if you have `module-resolver` in babelOptions.
3. The out-of-memory error. It can be okay if you have a massive code base because the memory usage is far from optimal right now, but if you have something like 100-200 components or even less, and it happens, it may be caused by some endless recursion in processing.

The first two are simple. The third one requires future investigation. First, I'm going to minimise memory usage.

Anber:
Memory-related work will be done here https://github.com/callstack/linaria/pull/1331

PierreGUI:
Thanks!  
FYI the MR fixed the multiple warnings, but I still get one when configuring `babelOptions` from Webpack:

`.babelrc`: 
```json
"plugins": [["module-resolver", { "alias": { "~": "./src" }}]
```

`webpack.js`:
```js
loader: '@linaria/webpack5-loader',
options: {
	babelOptions: {
		plugins: [
			["module-resolver", { "alias": { "~": "./src" }}],
		],
	},
```

PR #1334, opened 2023-09-05
Race condition in resolve

Anber:
## Motivation

Asynchronous resolver potentially could cause a race condition.

## Summary

Adding dependencies to an entrypoint has been moved from asynchronous code to synchronous.


Anber:
@PierreGUI we found and fixed something here. It might help with your build. Or might not :)

PierreGUI:
Thanks, I've tried this build, hélas unsuccessfully. The build of my big entrypoint gets stuck at about 10%, and then runs out of memory.

Note: prepare script fails for @linaria/rollup (but succeeded for babel)

Issue #1368, opened 2023-10-05
Invalid CSS: Space inserted after function interpolation


PierreGUI:
## Environment
- Linaria version: 5.0.2/5.0.3
- Bundler (+ version): Webpack 5

## Description
1. Use a function interpolation
2. Write CSS unit after the function
3. Use variable interpolation on the next line
:x: There's a space between value and unit (invalid CSS)

Example:
```jsx
export const FS = 12;
const getWidth = () => 10;

const H1 = styled.h1`
  width: ${getWidth}px;
  font-size: ${FS}px;
`;
```

Produces ->
<img width="228" alt="image" src="https://github.com/callstack/linaria/assets/1013904/8db729af-1a66-464d-b41b-1a6dd03bde78">

## Reproducible Demo
https://stackblitz.com/edit/linaria-bug-o9hmgr?file=src%2FTitle.tsx

layershifter:
@PierreGUI the repro is partially broken:

<img width="367" alt="image" src="https://github.com/callstack/linaria/assets/14183168/7166c8b2-c0ce-4bb4-8e92-25bf4d8b8032">

However, usage of the proper version does not fix the bug :) It's present.

---

The `transform()` returns in result following:

```
result.rules {
  '.hhnsy8v': {
    cssText: '\n  width: var(--hhnsy8v-0);\n  font-size: 12\n                  px;\n',
    className: 'hhnsy8v',
    displayName: 'H1',
    start: { line: 10, column: 11, index: 220 }
  }
}
```

I traced it to the following:

https://github.com/callstack/linaria/blob/e25e91ffb1187708680b97b06888adde7291b4d9/packages/tags/src/utils/templateProcessor.ts#L129-L130

- `toCSS()` returns `12` (`string`)
- `stripLines()` adds a newline that is considered as " "

This happens because:

https://github.com/callstack/linaria/blob/e25e91ffb1187708680b97b06888adde7291b4d9/packages/tags/src/utils/stripLines.ts#L14-L20

As `stripLisnes()` gets `{ start: { line: 11, column: 10 }, end: { line: 12, column: 18 } }` as `loc` input.

agriffis:
@layershifter I have not dug into this personally, just curious if you continued down this debugging journey to figure out why `stripLines` receives those different start/end lines. It seems like they should be the same, given the input?

agriffis:
Oh, here it is:

[...]

PR #1373, opened 2023-10-14
feat: upgrade stylis to v4

### LINKED ISSUE #649: Update stylis.js dependency to v4
## Describe the enhancement

Since 2.0 is coming, maybe it's not too late to update `stylis.js` dependency to v4.

## Motivation

There might be improvements in different areas as well as breaking changes.

## Possible implementations

Update `stylis.js` and hope tests pass 🙈 

## Related Issues

https://github.com/callstack/linaria/issues/582

agriffis:
## Motivation

Stylis v4 is faster than v3, and fixes misbehaviors in the older version regarding nested pseudo-classes.

Fixes #649 

## Summary

Update stylis in both the babel and atomic packages.

The changes are pretty straightforward. In the atomic package, rather than configure stylis keyframes/prefix, we simply omit those middlewares for the same effect.

This is a (slightly) breaking change, and the commit is marked as such. In particular, stylis v4 handles nested pseudo-classes differently from v3. The old version joined `.x { :is(.y) }` to `.x:is(.y)` whereas the new version separates them `.x :is(.y)`. The new behavior matches how SCSS and other preprocessors work. Linaria consumers will need to check and possibly update their styles for the new behavior.

## Test plan

Ran all the tests and updated snapshots:

- The babel snapshots updated in an innocuous way: the newer stylis includes a space where it didn't previously.

- The atomic snapshots updated in a way where something that was previously broken one way (`200;px`) is broken in a different way (`200 px`). Not a regression, so no reason to fix as part of this change.

### DIFF
--- a/.changeset/grumpy-flowers-watch.md
+++ b/.changeset/grumpy-flowers-watch.md
@@ -1,0 +1,7 @@
+---
+'@linaria/atomic': patch
+'@linaria/babel-preset': patch
+'@linaria/testkit': patch
+---
+
+Stylis has been upgraded from v3 to v4.
--- a/packages/atomic/package.json
+++ b/packages/atomic/package.json
@@ -56,12 +56,13 @@
     "@linaria/utils": "workspace:^",
     "known-css-properties": "^0.24.0",
     "postcss": "^8.4.31",
-    "stylis": "^3.5.4",
+    "stylis": "^4.3.0",
     "ts-invariant": "^0.10.3"
   },
   "devDependencies": {
     "@babel/types": "^7.22.15",
-    "@types/node": "^17.0.39"
+    "@types/node": "^17.0.39",
+    "@types/stylis": "^4.2.1"
   },
   "peerDependencies": {
     "react": ">=16"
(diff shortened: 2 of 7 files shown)


agriffis:
The atomic snapshot weirdness looks related to https://github.com/callstack/linaria/issues/1368

agriffis:
This will need test snapshots updated after merging #1374, then the atomic snapshot will no longer be broken.

Anber:
Thank you @agriffis!

agriffis:
@Anber I noticed in https://github.com/callstack/linaria/pull/1383 that this doesn't seem to be queued as a major version bump.

It doesn't matter to me personally what you do, but the stylis v4 upgrade changes how nested rules are parsed. It's a good change, in line with other processors, but it's a breaking change nonetheless.

What's your thinking here? Release as minor/micro and don't worry about it, or bump to v6?

Anber:
Yeah, you are right, it should be BC. Anyway, we are moving all transform and plugin logic to [another package](https://github.com/Anber/wyw-in-js). After that, Linaria will be just a custom processor on top of it. Your PR will be soon cherry-picked to wyw-in-js and released in wyw-in-js@1.0.0 and Linaria@6.0.0.

agriffis:
Interesting, thanks. I wasn't aware of wyw nor of Griffel. Looking forward to Linaria v6 :+1:

PR #1450, opened 2025-06-21
fix: update JSX type for old React support

nihgwu:
## Motivation

better version of https://github.com/callstack/linaria/pull/1420 which modified external namespace globally
we see error like `Property 'children' is missing in type '...' but required in type 'Props'. (ts 2741)` everywhere in a project using React 17, although it's working fine with projects using React 18 and React 19

### DIFF
--- a/.changeset/early-mice-deliver.md
+++ b/.changeset/early-mice-deliver.md
@@ -1,0 +1,5 @@
+---
+'@linaria/react': patch
+---
+
+Update types to support old React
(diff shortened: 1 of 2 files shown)


nihgwu:
@Anber Hi can you check, this is a blocker for us, otherwise we will need to lock to older version 🙏

nihgwu:
@thymikee sorry for bothering, but this is critical for us, can't find an easy solution from our side 🙏

thymikee:
Hey, I'm not maintaining this library, hope that @Anber can get to it when he's available :)

